PR #7, opened 2019-01-04
Added option for page numbers and fixed lang file twice

Vonr:
CommandBook now accepts /cbook <fill/random> <pages> on top of the old format /cbook <fill/random> which still simply makes 50 pages.


KaptainWutax:
:O

PR #128, opened 2020-07-01
Add cstacksize command

haykam821:
This pull request adds a new command to view the total number of stacks for a given item.

### DIFF
--- a/src/main/java/net/earthcomputer/clientcommands/ClientCommands.java
+++ b/src/main/java/net/earthcomputer/clientcommands/ClientCommands.java
@@ -43,6 +43,7 @@ public class ClientCommands implements ClientModInitializer {
         GlowCommand.register(dispatcher);
         GetDataCommand.register(dispatcher);
         ScriptCommand.register(dispatcher);
+        StackSizeCommand.register(dispatcher);
 
         CrackRNGCommand.register(dispatcher);
         
--- a/src/main/java/net/earthcomputer/clientcommands/command/StackSizeCommand.java
+++ b/src/main/java/net/earthcomputer/clientcommands/command/StackSizeCommand.java
@@ -1,0 +1,61 @@
+package net.earthcomputer.clientcommands.command;
+
+import com.mojang.brigadier.CommandDispatcher;
+
+import net.minecraft.client.MinecraftClient;
+import net.minecraft.item.ItemStack;
+import net.minecraft.server.command.ServerCommandSource;
+import net.minecraft.text.Text;
+import net.minecraft.text.TranslatableText;
+import net.minecraft.util.Hand;
+
+import static net.minecraft.command.arguments.ItemStackArgumentType.*;
+import static com.mojang.brigadier.arguments.IntegerArgumentType.*;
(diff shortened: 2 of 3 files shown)


Earthcomputer:
Just fyi, `/ccalc <n>` already displays the number of stacks and the remainder assuming the item stacks to 64. As for this PR, perhaps the command could have a better name?

zax71:
I think that would be a great idea!, where is a list of commands this mod adds?

OneYellowLemon:
Maybe /cstack

zax71:
i think either have the old one and /stacksize, just /stacksize or all 3

Earthcomputer:
The old one is a general purpose calculator that does maths, this is not a replacement for that. I don't like adding aliases because it encourages 12 year olds to ask "where are my 2 letter commands" like they don't know that tab exists. I'm happy to accept this PR only because it adds the extra functionality of calculating items that don't stack to 64, however I don't like the name /cstacksize, however I can't immediately think of a better name.

zax71:
thx for replying so quick. i think adding the finctionality would be GREAT! but i get what you meen

(ps i'm 12)

Earthcomputer:
Ah, sorry for generalizing, but I'm sure you know the types of players I mean ;)
How about `/ccalcstack`?

zax71:
yes, i do know who you mean and, i think that's a great name for the command, if it gets added soon i'l download the mod to add to my client side mod pack (for personal use).

so i don't download the wrong version of the mod, when will the update come out?

Earthcomputer:
I'll see about releasing a new version today.

zax71:
thank you if you could put a comment on here when you've added it so i get notified, thx for being so quick!

Earthcomputer:
Should be released now.

zax71:
ok, i'll download it!

zax71:
could you add it to the 1.15 version because i play on 1.15 because of optifine

Earthcomputer:
No. I don't maintain old versions. Also there are alternatives to optifine nowadays, like Sodium.

zax71:
oh, then i won't use it untill optifine updates sorry

Earthcomputer:
Sorry but optifine will be incompatible with 1.16 fabric, as we're dropping support for it. It's ancient, not even as effective as Sodium, and causes tons of mod incompatibilities which are hard to fix because the owner is a dick and doesn't care about compatibility.

zax71:
ok, i'll try sodium sometime

zax71:
can you give me a download link to sodium, i can't find a .jar file

seanguy244:
https://github.com/jellysquid3/sodium-fabric repo for the mod

zax71:
Where’s the .jar tho? I can’t find it can you give me a more direct link

Earthcomputer:
Google "fabric sodium download" second result for me.

zax71:
It’s just the github page that shows up can you give me a link to the downloads page?

Earthcomputer:
It's on curseforge. Now this has got very off topic.

PR #131, opened 2020-07-15
Update branch_mine.js for 1.16 materials

illumincrotty:
add netherite to the material list


Earthcomputer:
It's only an example script and not part of clientcommands, but sure why not.

PR #133, opened 2020-07-23
Added Polish Language support,Updated en_us.json & pl_pl.json

L0um15:
I also edited BookCommand to make "You are not holding a book" translatable to other languages and removed unused import.
There is also a word that i cant translate because of many meanings. In the future i will update it.

### DIFF
--- a/src/main/resources/assets/clientcommands/lang/pl_pl.json
+++ b/src/main/resources/assets/clientcommands/lang/pl_pl.json
@@ -1,0 +1,123 @@
+{
+  "commands.cbook.success": "Pomyślnie zedytowano książkę",
+
+  "commands.ccalc.expected": "Spodziewane %s",
+  "commands.ccalc.invalidArgumentCount": "Funkcja \"%s\" nie może mieć %d argumentów",
+
+  "commands.ccalcstack.success": "%d %s to %d staki/ów i %d dodatkowe",
+  "commands.ccalcstack.success.empty": "%d przedmiotow to %d staki/ów i %d dodatkowe",
+  "commands.ccalcstack.success.empty.exact": "%d przedmiotów to dokładnie %d staków",
+  "commands.ccalcstack.success.exact": "%d %s to dokładnie %d staki",
+
+  "commands.ccrackrng.starting": "Łamanie ziarna gracza",
+  "commands.ccrackrng.success": "RNG gracza złamane: %d",
+
+  "commands.cenchant.expectedWithWithout": "Spodziewane \"with\"/\"without\"",
+  "commands.cenchant.failed": "To niemożliwe albo zajmie to zbyt długo aby otzymać te zaklęcia",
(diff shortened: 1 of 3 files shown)


src/main/resources/assets/clientcommands/lang/pl_pl.json
```diff
@@ -0,0 +1,124 @@
...
+  "enchCrack.xpSeed.one": "Ziarno XP: %08X",
+  "enchCrack.xpSeed.many": "Liczba możliwych ziarn XP: %d",
+  "enchCrack.playerRNGSeed.many": "Liczba możliwych RNG gracza: %d",
+  "enchCrack.addInfo": "Dodaj Informacje",
+
+  "enchCrack.insn.dummy": "Zaklnij zbędny przedmiot",
+  "enchCrack.insn.ready": "Twoje ziarno zaklęcia jest gotowe",
+  "enchCrack.insn.bookshelves": "Półki na książki: %d",
+  "enchCrack.insn.slot": "W Slocie: %d",
+  "enchCrack.insn.itemThrows": "Ile musisz wyrzucić przedmiotów: %d (about %f seconds)",
```
Earthcomputer: Could `about %f seconds` also be translated?

src/main/resources/assets/clientcommands/lang/pl_pl.json
```diff
@@ -0,0 +1,124 @@
...
+  "enchCrack.insn.itemThrows.noDummy": "Nie potrzebujesz zbędnego przedmiotu",
```
Earthcomputer, line 81: Google translate tells me that "zbędnego" means "unnecessary". The English is more trying to hint at a "throwaway" item more than an unnecessary one. Putting "throwaway" into Google translate and looking at the adjective, it's giving me "do jednorazowego użytku". Would that be more suitable?

L0um15: Polish language is very difficult to translate and as i said we have many meanings. "do jednorazowego użytku" means "one time use" and "zbędnego" means that you're not gonna use that enchanted item and wanna disenchant that in order to use it again as a dummy. Im not saying that "do jednorazowego użytku" does not fit at all. But i think that "zbędnego" fits perfectly. Short Answer: Both are correct.

src/main/resources/assets/clientcommands/lang/pl_pl.json
L0um15: I didn't saw that. I edited it now.

src/main/resources/assets/clientcommands/lang/pl_pl.json
L0um15: Zbędny also means something with no value as wooden_sword

Earthcomputer: I'll take your word for it, just checking

Earthcomputer (approved):
Ty!

PR #135, opened 2020-08-03
Improved DurabilityMessage visibility in bright areas, Added sound hint when tool durability is below 15, Added cgamma translation

L0um15:


Earthcomputer (approved):
Ah, sorry, I looked at this and then forgot about it

PR #142, opened 2020-08-24
Create zh_cn.json

User670:
(If you wonder why space after a full-width punctuation, ... Minecraft font is broken, and the default Unicode font displays most full-width punctuation marks - other than the full stop - as half-width.)

### DIFF
--- a/src/main/resources/assets/clientcommands/lang/zh_cn.json
+++ b/src/main/resources/assets/clientcommands/lang/zh_cn.json
@@ -1,0 +1,126 @@
+{
+  "commands.cbook.success": "成功编辑书本",
+  "commands.cbook.commandException": "您没有手握书本",
+
+  "commands.ccalc.expected": "期望值 %s",
+  "commands.ccalc.invalidArgumentCount": "函数 \"%s\" 无法接受 %d 个自变量",
+
+  "commands.ccalcstack.success": "%d 个 %s 是 %d 组余 %d 个",
+  "commands.ccalcstack.success.empty": "%d 个物品是 %d 组余 %d 个",
+  "commands.ccalcstack.success.empty.exact": "%d 个物品是正好 %d 组",
+  "commands.ccalcstack.success.exact": "%d 个 %s 是正好 %d 组",
+
+  "commands.ccrackrng.starting": "正在破解玩家随机种子",
+  "commands.ccrackrng.success": "玩家随机种子已破解： %d",
+
+  "commands.cenchant.expectedWithWithout": "有效参数： \"with\" 或 \"without\"",
+  "commands.cenchant.failed": "您请求的附魔无法获得或者需要太长时间",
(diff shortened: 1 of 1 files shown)


src/main/resources/assets/clientcommands/lang/zh_cn.json
```diff
@@ -0,0 +1,126 @@
...
+  "playerManip.state.cracked": "已破解",
+  "playerManip.state.ench_cracking_1": "正在破解附魔种子 1",
+  "playerManip.state.half_cracked": "半破解",
+  "playerManip.state.ench_cracking_2": "正在破解附魔种子 2",
+  "playerManip.state.cracking": "正在破解",
+  "playerManip.state.eating": "正在吃食物",
+  "playerManip.state.manipulating_enchantments": "正在调整附魔",
+  "playerManip.state.waiting_dummy_enchant": "等待附魔其他物品",
+
+  "playerManip.reset": "重新开始调整玩家随机数。原因： %s",
```
Earthcomputer: I'm aware that google translate can be inaccurate, so please let me know if I'm wrong.
But I think "调整" translates to "adjust", whereas the English version is "manipulate". In the context of RNG manipulation, wouldn't "操作" or maybe "操纵" make more sense?

User670: Ah, thanks. I am not aware of an equivalent term of "RNG manipulation" in Chinese, so I borrowed the term from Japanese. I guess I'll do some more digging to see what people call "RNG manipulation"...

User670:
RNG "manipulation" now uses the word 操控.

Actually I have a question, what is XP seed supposed to be? Experience seed? (does experience need a seed? Or is it actually enchantment seed?)

Earthcomputer:
XP seed and enchantment seed are just 2 phrases that mean the same thing. The player seed generates the enchantment seeds.

src/main/resources/assets/clientcommands/lang/zh_cn.json
```diff
@@ -0,0 +1,126 @@
...
+
+  "playerManip.state": "玩家种子破解状态： %s",
+  "playerManip.state.uncracked": "未破解",
+  "playerManip.state.cracked": "已破解",
+  "playerManip.state.ench_cracking_1": "正在破解附魔种子 1",
+  "playerManip.state.half_cracked": "半破解",
+  "playerManip.state.ench_cracking_2": "正在破解附魔种子 2",
+  "playerManip.state.cracking": "正在破解",
+  "playerManip.state.eating": "正在吃食物",
+  "playerManip.state.manipulating_enchantments": "正在调整附魔",
```
Earthcomputer: You missed one, but otherwise looks good ;)

Earthcomputer (changes requested):

Earthcomputer (approved):

Earthcomputer:
Ty!

PR #158, opened 2020-11-30
Make /ccheatcrackrng return old seed if existent for testing

Vonr:
Makes /ccheatcrackrng print "Player RNG cracked: %oldSeed% -> %newCheatCrackedSeed%" 
This is useful for testing rng advancement.


Earthcomputer (approved):
Good idea

ImUrX:
THANKS GOD!

PR #167, opened 2021-01-05
Added higlighting(rendering) system.

mariusdkm:
Draw/highlight what ever you want.

Changed cglow to:
`/cglow (block|entity) ...`

Fixed some small bugs in ChorusManipulation.  
Goal is now rendered.
Made instructions more clear.

I'm not entirely sure about the strucure of the renderin system.
Because there isn't one centralised rendering queue.  
Instead you register in `ClientCommands.java` your own function,  
which is will handle the rendering.
The upside is one can inject renders in multiple places along the rendering pipeline  
and have their own rendering pipeline customized.
~~The downside is, it's hard to keep track of all the renders?~~  
Base structure is by KaptainWutax in his SeedCracker mod.
[https://github.com/KaptainWutax/SeedCracker](https://github.com/KaptainWutax/SeedCracker) 
Though I changed a lot.

I haven't look into it, but I think it would be usefull to be able to draw things using scripts.

I hope I didn't break anything with my merge...


Earthcomputer:
Please can you split this into multiple PRs, one for each change? Thanks.

PR #168, opened 2021-01-05
Render

mariusdkm:
Draw/highlight what ever you want.
Well theoretically, no function to do that from ingame is wirtten for that.
Highlighting see next pr...

- Fixed some small bugs in ChorusManipulation.  
- Goal is now rendered.
- Made instructions when using `/cchorus` more clear.

grafik

I'm not entirely sure about the structure of the rendering system.
Because there isn't one centralised rendering queue.  
Instead you register in `ClientCommands.java` your own function,  
which is will handle the rendering.
The upside is one can inject renders in multiple places along the rendering pipeline  
and have their own rendering pipeline customized.
~~The downside is, it's hard to keep track of all the renders?~~  
Base structure is by KaptainWutax in his SeedCracker mod.
[https://github.com/KaptainWutax/SeedCracker](https://github.com/KaptainWutax/SeedCracker) 
Though I changed a lot.

I haven't look into it, but I think it would be usefull to be able to draw things using scripts.


src/main/java/net/earthcomputer/clientcommands/features/ChorusManipulation.java
```diff
@@ -11,30 +16,56 @@
 import net.minecraft.util.math.Vec3d;
 import net.minecraft.world.World;
 
-import static net.earthcomputer.clientcommands.command.ClientCommandManager.sendError;
-import static net.earthcomputer.clientcommands.command.ClientCommandManager.sendFeedback;
+import java.awt.*;
```
Earthcomputer: Don't use AWT, it crashes on MacOS with LWJGL. Use a packed int colour instead.

src/main/java/net/earthcomputer/clientcommands/render/Renderer.java
```diff
@@ -0,0 +1,43 @@
...
+     * Nothing
+     */
+    protected int life = -1;
+    protected long prevTick;
+
+    public void render() {
+        if (this.mc.world.getTime() != this.prevTick && this.life > 0) {
+            this.life--;
+            this.prevTick = this.mc.world.getTime();
+        }
```
Earthcomputer:
```suggestion
        if (this.life > 0) {
            this.life -= this.mc.world.getTime() - this.prevTick;
        }
```
Fixes the situation where there are multiple ticks per frame.

src/main/java/net/earthcomputer/clientcommands/render/Renderer.java
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    public abstract BlockPos getPos();
+
+    public int getLife() {
+        return this.life;
+    }
+
+    public boolean shouldKill() {
```
Earthcomputer: Where is this called?

src/main/java/net/earthcomputer/clientcommands/mixin/MixinDummyProfiler.java
```diff
@@ -0,0 +1,18 @@
...
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+
+@Mixin(DummyProfiler.class)
+public class MixinDummyProfiler {
+
+    @Inject(method = "swap(Ljava/lang/String;)V", at = @At("HEAD"))
+    private void swap(String type, CallbackInfo ci) {
+        RenderQueue.get().onRender(type);
```
Earthcomputer: This looks a little inefficient and very unreliable. There could be multiple times the profiler gets called in different places with the same string. It would be better to have an enum render layer and inject in a set number of places. That would also allow you to pass `delta` so the rendering doesn't look jittery and also you wouldn't need to store the matrix stacks in a weird way.

Earthcomputer (changes requested):
Sorry, but the overall quality of the code you have taken is awful, and needs to be completely rewritten

src/main/java/net/earthcomputer/clientcommands/render/Renderer.java
mariusdkm: I had this for cglow, to check if one object should still be renderd.
I my though was I should centralise it. 
```java
for (Iterator<Renderer> iterator = renderers.iterator(); iterator.hasNext(); ) {
    Renderer renderer = iterator.next();
    renderer.render();
    if (renderer.shouldKill()) {
        // Remove the current element from the iterator and the list.
        iterator.remove();
    }
}
```

Mhowser:
Sorry to disturb if I am, but would it be possible to add an RGB color
option? Something that cycles through the rainbow colors in order to
indicate an entity or block of significance. I think that would be really
helpful.

Mhowser:
I just thought of a better idea.

It would be extremely handy if the command /cfind had a --glow option. An
example could be `/cfindblock minecraft:spawner --glow RGB 30`

Earthcomputer:
@Mhowser this PR doesn't have anything to do with `/cglow`. Please create a separate issue for this feature request, where it won't get lost.

Mhowser:
Alright.

mariusdkm:
Ok I tried to include all of your requested changes.

> It would be better to have an enum render layer and inject in a set number of places. 

I did that and also changed the whole renderQueue to static instead of having an instance object.
I have injected into two places along the rendering pipeline, although I only use one of them.
I didn't test out, what different injection places result in.

I haven't implemented anything with the delta yet, since I don't know how to use it...

I also noticed, that when the game was lagging the rng wouldn't sometimes break after one or two chorus fruit teleportations. I've fixed that as good as I can, sometimes in rare cases the rng still breakes.
I'm no longer changing the rng after the item should be consumed.
I'm now also accounting for this in the rng maintenance.
I hope the fix works...

src/main/java/net/earthcomputer/clientcommands/render/RenderQueue.java
```diff
@@ -0,0 +1,55 @@
...
+import net.minecraft.client.util.math.MatrixStack;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.function.Consumer;
+
+public class RenderQueue {
+
+    private static final HashMap<String, List<Consumer<MatrixStack>>> queue = new HashMap<>();
```
Earthcomputer: The key to this map could now simply be an InjectLoc rather than String. And if you're doing that, may as well change it from a HashMap to an EnumMap.

src/main/java/net/earthcomputer/clientcommands/mixin/MixinGameRenderer.java
```diff
@@ -0,0 +1,23 @@
...
+
+    @Inject(method = "renderWorld", at = @At(value = "INVOKE", target = "Lnet/minecraft/util/profiler/Profiler;swap(Ljava/lang/String;)V"))
+    private void renderWorldHand(float delta, long time, MatrixStack matrixStack, CallbackInfo ci) {
+        RenderQueue.InjectLoc.HAND.onRender(delta, time, matrixStack);
+    }
+
+    @Inject(method = "renderWorld", at = @At(value = "INVOKE", target = "Lnet/minecraft/util/profiler/Profiler;swap(Ljava/lang/String;)V"))
+    private void renderWorldCamera(float delta, long time, MatrixStack matrixStack, CallbackInfo ci) {
+        RenderQueue.InjectLoc.CAMERA.onRender(delta, time, matrixStack);
+    }
```
Earthcomputer: Don't these both inject into the same location? I think INVOKE_STRING might be what you're looking for.

src/main/java/net/earthcomputer/clientcommands/ClientCommands.java
```diff
@@ -22,6 +22,8 @@ public void onInitializeClient() {
         //noinspection ResultOfMethodCallIgnored
         configDir.mkdirs();
 
+        RenderQueue.add(RenderQueue.InjectLoc.HAND, ChorusManipulation::renderChorusGoal);
```
Earthcomputer: What I was more thinking, is you add shapes from the tick loop, in amongst your other code. It's more convenient that way because you'll have access to all the temporary variables you're using. You add the shapes themselves to the render queue, and they'll then be rendered in the render loop.

Earthcomputer (changes requested):
Ignore what I said about the delta, that's only relevant for moving objects, your shapes don't move.

Since I think we're getting closer to the solution I was imagining, if you need help implementing these changes, message me on discord and I'll try and find time to help.

src/main/java/net/earthcomputer/clientcommands/mixin/MixinGameRenderer.java
mariusdkm: Oh yea, you're right... missed that

src/main/java/net/earthcomputer/clientcommands/ClientCommands.java
mariusdkm: I'm not sure about that, since the way it's at the moment,  
if somebody decides to add something to the render queue in the future,
they can add their own render pipeline.
There they than can do whatever they want.
Here for the cglow command I had updated, the cglow had it's own renderFunction
https://github.com/Earthcomputer/clientcommands/pull/167/files#diff-49d93d00b6c6b62ab7e077685034fb2dd58e70fc76174acb59aea70e5152ed1aR134

Earthcomputer (approved):
I have made some improvements. It's hard to see the advantage just from the chorus manipulation, but doing it this way round will benefit a lot of the other code that wants to use this system a lot.

PR #170, opened 2021-01-09
Silence fishing rod sounds when manipulating fishing

DJtheRedstoner:
It doesn't silence 100% of the sounds but it silences the very annoying rod casting sound. Not worked on fabric mods in a very long to so if there is a better way to do this please let me know.


Earthcomputer:
Honestly I'd prefer if this was a config option disabled by default. Waiting on Quilt's config API.

xpple:
This would be closed by the implementation of #466 right?

PR #180, opened 2021-02-12
Added /cspeed, /cdab, /cclip, and /cvclip

Mr-Sbev:


Mr-Sbev:
I'm sorry, but I couldn't figure out how to limit the speed change to only creative, or non elytra. It just uses velocity changes, and can be a bit strange at times. It also takes a multiplier but one isn't normal walking speed...

Earthcomputer:
I'm sorry, but there are many problems with this PR. I said to make one because you said you had already written the code. However, three of the commands are out of the scope of clientcommands, and the one that isn't out of scope (cspeed for creative mode) is incomplete, appears difficult to maintain (a better way to do it would be to add an attribute modifier). Also, there are many style issues throughout, where you are not following the code style of code elsewhere in the project.

Mr-Sbev:
Should I continue to try to implement cspeed? I couldn't figure out how to use an attribute modifier, but I'm new to modding Minecraft so I probably missed something. Are there any specific examples of ways that I could improve the style?

Mr-Sbev:
Also, I assume that by the out of scope commands, you were referring to /cdab, /cclip, and /cvclip... Are they too cheaty?

PR #183, opened 2021-02-15
Updated /cglow to allow blocks and areas to be highlighted

mariusdkm:
The title says it...

I'm unsure about two things:
* Whether `MinecraftClient.getInstance().world` (GlowCommand.java, L130)
  is the right way to get the world in the command context.
* My "fix" in order to allow rendering through blocks is right (MixinGameRenderer.java, L28)  
   `RenderSystem.clear(256, MinecraftClient.IS_SYSTEM_MAC);`

I intentionally didn't add the ability to use a `--keep-searching` flag to `/cglow blocks` and made the input only coordinates,  
since there is already a command, that finds blocks for somebody named `/cfindblock`.

I instead made the output of `/cfindblock` "glowable" instead of "lookable."  
One could also add a flag to `/cfindblock` to automatically glow the found block...

### DIFF
--- a/src/main/java/net/earthcomputer/clientcommands/mixin/MixinGameRenderer.java
+++ b/src/main/java/net/earthcomputer/clientcommands/mixin/MixinGameRenderer.java
@@ -2,6 +2,7 @@ package net.earthcomputer.clientcommands.mixin;
 
 import com.mojang.blaze3d.systems.RenderSystem;
 import net.earthcomputer.clientcommands.render.RenderQueue;
+import net.minecraft.client.MinecraftClient;
 import net.minecraft.client.render.BufferBuilderStorage;
 import net.minecraft.client.render.Camera;
 import net.minecraft.client.render.GameRenderer;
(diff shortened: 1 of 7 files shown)


src/main/java/net/earthcomputer/clientcommands/mixin/MixinGameRenderer.java
```diff
@@ -22,11 +23,16 @@
     @Inject(method = "renderWorld", at = @At(value = "INVOKE_STRING", target = "Lnet/minecraft/util/profiler/Profiler;swap(Ljava/lang/String;)V", args = {"ldc=hand"}))
     private void renderWorldHand(float delta, long time, MatrixStack matrixStack, CallbackInfo ci) {
         matrixStack.push();
+
+        //Render lines through everything
+        RenderSystem.clear(256, MinecraftClient.IS_SYSTEM_MAC);
```
Earthcomputer: Not sure what 256 is, can you check if this is `GL11.GL_DEPTH_BUFFER_BIT`? Always used named constants in GL calls rather than magic constants like this.

Earthcomputer (changes requested):
Not sure if this is the best approach to rendering through blocks but it'll do for now. Maybe add a "// TODO: is this the best approach to render through blocks?" comment

Earthcomputer (approved):
Looks good, thank you!

PR #189, opened 2021-02-21
Allow for namespaced commands and move commands to a separate package

xpple:
By using `reader.readUnquotedString()` to parse the string, commands cannot contain `:`. By allowing commands with `:`, one can create pseudo namespaced commands (vanilla Minecraft does not support properly namespaced commands). This can be useful to prevent conflicts between client-sided and server-sided literals.
I chose not to Mixin `StringReader` and change the method because that could have unwanted side-effects. For this reason I created a subclass of `StringReader`, which inherits everything from `StringReader` but the regarding method.

Furthermore, I moved all the commands to a new package: `command.commands`, which makes the file structure a lot more clear. Before, classes like `ClientCommandManager` were hard to find as they were a needle in a haystack of command classes.

Any suggestions are welcome.

### DIFF
--- a/src/main/java/net/earthcomputer/clientcommands/ClientCommands.java
+++ b/src/main/java/net/earthcomputer/clientcommands/ClientCommands.java
@@ -1,9 +1,8 @@
 package net.earthcomputer.clientcommands;
 
 import com.mojang.brigadier.CommandDispatcher;
-import net.earthcomputer.clientcommands.command.*;
-import net.earthcomputer.clientcommands.features.ChorusManipulation;
-import net.earthcomputer.clientcommands.render.RenderQueue;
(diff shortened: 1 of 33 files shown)


mariusdkm:
Wouldn't it also be advantageous to have a somewhat unified command class name by removing the `C`'s in front the few commands, that still have it, like `CEnchantCommand`?

Earthcomputer:
I'm not planning on adding any colons in my commands

xpple:
> I'm not planning on adding any colons in my commands

I see. Don't you think it's cleaner to have `cc:enchant` instead of `cenchant` though?

What are your thoughts on the file structure commit?

I must also agree with mariusdkm, consistency would be nice.

PR #201, opened 2021-03-11
Move all commands to separate package

xpple:
I'm surprised no one has done this before me. I feel like this improves the project structure and tidies up the `command` package.


Earthcomputer:
> I feel like this improves the project structure and tidies up the command package.

That's just a feeling, I don't share it, plus it doesn't reflect the vanilla package structure.

PR #209, opened 2021-03-20
Add the cfov command

### LINKED ISSUE #203: [Suggestion] /cfov?
So we can zoom in / out without of limitation in the game? (minimal 30, maximum 110) so the syntax probably very simple `/cfov <fov>`

haykam821:
This pull request fixes #203 by adding the `/cfov` command. This command allows setting the FOV to any double value, in addition to the vanilla presets, `normal` (70) and `quakePro` (110).

### DIFF
--- a/src/main/java/net/earthcomputer/clientcommands/command/FovCommand.java
+++ b/src/main/java/net/earthcomputer/clientcommands/command/FovCommand.java
@@ -18,14 +18,14 @@ public class FovCommand {
 
         dispatcher.register(literal("cfov")
             .then(argument("fov", doubleArg())
-                .executes(ctx -> setGamma(ctx.getSource(), getDouble(ctx, "fov"))))
+                .executes(ctx -> setFov(ctx.getSource(), getDouble(ctx, "fov"))))
             .then(literal("normal")
-                .executes(ctx -> setGamma(ctx.getSource(), 70)))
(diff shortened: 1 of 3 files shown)


src/main/java/net/earthcomputer/clientcommands/command/FovCommand.java
```diff
@@ -0,0 +1,37 @@
...
+        dispatcher.register(literal("cfov")
+            .then(argument("fov", doubleArg())
+                .executes(ctx -> setGamma(ctx.getSource(), getDouble(ctx, "fov"))))
+            .then(literal("normal")
+                .executes(ctx -> setGamma(ctx.getSource(), 70)))
+            .then(literal("quakePro")
+                .executes(ctx -> setGamma(ctx.getSource(), 110))));
+    }
+
+    private static int setGamma(ServerCommandSource source, double fov) {
```
Earthcomputer: setFov

Earthcomputer (changes requested):

src/main/java/net/earthcomputer/clientcommands/command/FovCommand.java
haykam821: Fixed in f2cde2e0ecb5c6bc39483b47a5b6153d8b0898fc

Earthcomputer (approved):
Thanks!

PR #210, opened 2021-03-20
Remove an unnecessary redirector in the MixinEntityRenderDispatcher class

haykam821:


src/main/java/net/earthcomputer/clientcommands/mixin/MixinEntityRendererDispatcher.java
```diff
@@ -23,9 +19,11 @@ public void onConfigure(World world, Camera camera, Entity entity, CallbackInfo
         RenderSettings.preRenderEntities();
     }
 
-    @Redirect(method = "shouldRender", at = @At(value = "INVOKE", target = "Lnet/minecraft/client/render/entity/EntityRenderer;shouldRender(Lnet/minecraft/entity/Entity;Lnet/minecraft/client/render/Frustum;DDD)Z"))
-    public boolean redirectShouldRender(EntityRenderer<Entity> entityRenderer, Entity entity, Frustum frustum, double x, double y, double z) {
-        return entityRenderer.shouldRender(entity, frustum, x, y, z) && RenderSettings.shouldRenderEntity(entity);
+    @Inject(method = "shouldRender", at = @At("RETURN"), cancellable = true)
+    public void redirectShouldRender(Entity entity, Frustum frustum, double x, double y, double z, CallbackInfoReturnable<Boolean> ci) {
+        if (!RenderSettings.shouldRenderEntity(entity)) {
```
Earthcomputer: `RenderSettings.shouldRenderEntity` can be expensive in some cases, so this should check if `ci.getReturnValueZ()` is already false before calling that.

Earthcomputer (changes requested):
Oh lol, thanks :)

src/main/java/net/earthcomputer/clientcommands/mixin/MixinEntityRendererDispatcher.java
haykam821: Fixed in f40e02d80d8d0291b199f13d4ed9e21f5acf7623

Earthcomputer (approved):
Thanks!

PR #217, opened 2021-03-21
Fix Command Redirects to avoid duplication of optional parameters

### LINKED ISSUE #164: You can do "cenchant --simulate --simulate" when "--simulate" should only be allowed once
Tested on Client Commands 2.4.8 for Minecraft 1.16.4

Command and example output (actual):
image

Expected: Not to allow `--simulate` flag twice

Domkeykong:
This fixes #164

### DIFF
--- a/src/main/java/net/earthcomputer/clientcommands/command/CEnchantCommand.java
+++ b/src/main/java/net/earthcomputer/clientcommands/command/CEnchantCommand.java
@@ -23,11 +23,9 @@ public class CEnchantCommand {
         addClientSideCommand("cenchant");
 
         dispatcher.register(literal("cenchant")
+                .then(itemAndEnchantmentsPredicateArgument(false))
                 .then(literal("--simulate")
-                        .then(itemAndEnchantmentsPredicateArgument(true)))
-                .then(
-                        itemAndEnchantmentsPredicateArgument(false)
-                ));
+                        .then(itemAndEnchantmentsPredicateArgument(true))));
     }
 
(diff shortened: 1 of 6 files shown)


Earthcomputer (changes requested):
There's a couple of things I don't like this particular fix. The main thing is that it doesn't scale well. Adding an extra flag will double the amount of lines, so I'd prefer a solution that automatically generates the command tree for the flags, given a list of flags. However, even that doesn't solve the memory issue.
You have also inconsistently kept to use of withFlags. I like how you have removed the flags in FindItemCommand, as flags were a hacky workaround in the first place.
Also minor nitpick, please only put newlines before periods (.) in commands, and follow the formatting of other code in the project.

I'm wondering if it would be possible to have a fix that simply doesn't suggest the same flag again for auto-completion, even if they would technically be syntactically valid.

Domkeykong:
Currently there isn't really a method of providing default arguments!
This was already discussed by you @Earthcomputer in Mojang/brigadier#18 / Mojang/brigadier#19 and an implementation by @NeunEinser Mojang/brigadier#42 remains to be merged

Earthcomputer:
I still think changing the suggestor is the best solution, but I'll keep this PR open in case I change my mind.
(You missed a couple of withFlags btw, the best way to find them is to delete the withFlags and getFlags methods entirely and see if the project compiles).

Earthcomputer:
Have you pulled recently? There is also one in /ccalc that was added a couple of days ago.

Earthcomputer (approved):
As a fix this looks fine for now. Not going to merge just yet as I still want to see if I can find a solution I'm happier with.

PR #229, opened 2021-04-10
Allow validating open water for fishing bobbers

haykam821:
This pull request adds the `/cfish validate-open-water` subcommand, which determines if the player's current fishing bobber is considered to be in open water and thus able to fish up treasure.


xpple:
Suggestion: 

* shorten the subcommand literal to `validate`
* shorten the feedback to _something shorter_

haykam821:
If the intent is to have short subcommands, then the other ones should be renamed as well.

xpple:
I mean - doesn't `validate` say enough?

haykam821:
No, there are a lot of other things for fishing bobbers that can be validated.

xpple:
Would those be nice additions as well? If so, a general `validate` subcommand would make sense to have

Earthcomputer:
The problem with this check as implemented is that the fishing bobber not only needs to be in open water, but it needs to have always have been in open water from the moment it enters the water.

haykam821:
Is that how open water works in vanilla? I had trouble understanding the decompiled logic.

The intent of the command is that one can verify the setup without having to re-cast the fishing rod each time. I can add a third case to distinguish between the position being considered open water in general or just open water for the current fishing bobber.

Earthcomputer:
Yes that's how open water works in vanilla. If you look in `FishingManipulator` there is code there to simulate the fishing bobber before it's thrown, you could probably use that.

haykam821:
Why would it need to be simulated? Shouldn't it only need to check the actual `inOpenWater` field of the fishing bobber?

Earthcomputer:
Good point

Earthcomputer:
This still needs to be changed to use the `inOpenWater` field.

haykam821:
On 1.18.1, the client no longer updates `inOpenWater`. Would it be more trouble than it's worth to simulate the fishing bobber without fishing cracking? The message can still be changed to indicate that the fishing bobber must always be in open water for treasure.

Earthcomputer:
Superseded by https://github.com/Earthcomputer/clientcommands/commit/6640836cee2243142a24cefe67c08f143eec6955

haykam821:
This functionality would still be useful for validating open water in situations that do not involve fishing manipulation.

PR #230, opened 2021-04-11
Greatly improve efficiency of IO

xpple:
I wanted to do the IO this way earlier, but that wasn't possible as I was storing `PlayerInventory`s which couldn't be loaded without the player. Now that `ListTag`s are stored instead (as suggested by Earthcomputer 👍 ), this **is** possible. The file is only loaded once `onInitializeClient`, and will only be saved afterwards. If the file couldn't be read, the command is logically not registered.

### DIFF
--- a/src/main/java/net/earthcomputer/clientcommands/command/KitCommand.java
+++ b/src/main/java/net/earthcomputer/clientcommands/command/KitCommand.java
@@ -12,6 +12,7 @@ import net.fabricmc.fabric.api.util.NbtType;
 import net.fabricmc.loader.api.FabricLoader;
 import net.minecraft.SharedConstants;
 import net.minecraft.client.MinecraftClient;
+import net.minecraft.client.resource.language.I18n;
 import net.minecraft.command.CommandSource;
 import net.minecraft.datafixer.TypeReferences;
 import net.minecraft.entity.player.PlayerInventory;
@@ -21,6 +22,8 @@ import net.minecraft.screen.slot.Slot;
 import net.minecraft.server.command.ServerCommandSource;
 import net.minecraft.text.TranslatableText;
 import net.minecraft.util.Util;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 import java.io.File;
 import java.io.IOException;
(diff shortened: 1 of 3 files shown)


haykam821:
Why not load it the first time it is required?

xpple:
Funnily I also had that at one point, but I didn’t like having to set a Boolean to false in 5 methods just to delay the IO of one file

xpple:
More importantly, if you do it the first time it is required, the `suggest`ions won’t be available because the file and thus `Map` hasn’t been loaded

xpple:
> More importantly, if you do it the first time it is required, the `suggest`ions won’t be available because the file and thus `Map` hasn’t been loaded

This PR also fixes that issue btw

Earthcomputer (changes requested):
I don't like calling this directly inside `ClientCommands`, it breaks encapsulation. Probably best inside a `static {}` block.

xpple:
I thought it’d make sense to have it here, because it’s where all the commands are registered. Similar to the one which is being registered based on an `if`, this one is registered based on the fact whether the IO worked or not. Surely you don’t want to register the command inside the command’s class?

Earthcomputer:
Why would the I/O not work? That's an error case, not a simple if "either-could-happen" case. Just log something to the console to be honest.

xpple:
Honestly no idea how (or why) the IO would not work, but then again, it’s an _exception_! And I _am_ logging something to the console if it doesn’t work. I extracted the error handling to the `ClientCommands` class, because depending on the success of the IO, the command will be registered (or not). If the IO didn’t work the entire command is useless, and therefore useless to have registered.

src/main/java/net/earthcomputer/clientcommands/command/KitCommand.java
```diff
@@ -48,6 +50,15 @@
 
     private static final Map<String, ListTag> kits = new HashMap<>();
 
+    static {
+        try {
+            loadFile();
+        } catch (IOException e) {
+            final Logger logger = LogManager.getLogger();
```
Earthcomputer: Loggers should be a private static final field at the top of the class, and should specify the name "clientcommands".

src/main/java/net/earthcomputer/clientcommands/command/KitCommand.java
```diff
@@ -48,6 +50,15 @@
...
+            logger.info(I18n.translate("commands.ckit.loadFile.failed"));
```
Earthcomputer: Log messages don't need to be translated.

Earthcomputer (changes requested):

Earthcomputer (approved):
Lgtm

PR #232, opened 2021-04-13
Add /ckit preview <name>

xpple:
Few things. I decided to create my own `PreviewScreen` class, mainly because a `PlayerInventoryScreen` can only be instantiated with a player entity, which is not worth creating. Doing it this way also gets rid of slot interaction code, which might de-sync the client. Even with my own `PreviewScreen` class, you cannot use the player's `PlayerScreenHandler` because that is what actually contains the player's items, where it should contain the kit's items to display. For this reason a `new PlayerScreenHandler()` had to be instatiated. The opening of the screen itself is done using `client.send()`, which _I find_ to be cleaner than doing something like this:
```java
ClientTickEvents.START_CLIENT_TICK.register(client -> {
    if (KitCommand.openScreen != null) {
        client.openScreen(KitCommand.openScreen);
        KitCommand.openScreen = null;
    }
});
```

### DIFF
--- a/src/main/java/net/earthcomputer/clientcommands/command/KitCommand.java
+++ b/src/main/java/net/earthcomputer/clientcommands/command/KitCommand.java
@@ -1,5 +1,6 @@
 package net.earthcomputer.clientcommands.command;
 
+import com.mojang.blaze3d.systems.RenderSystem;
 import com.mojang.brigadier.CommandDispatcher;
 import com.mojang.brigadier.arguments.StringArgumentType;
 import com.mojang.brigadier.exceptions.CommandSyntaxException;
@@ -12,14 +13,20 @@ import net.fabricmc.fabric.api.util.NbtType;
 import net.fabricmc.loader.api.FabricLoader;
 import net.minecraft.SharedConstants;
 import net.minecraft.client.MinecraftClient;
+import net.minecraft.client.gui.screen.ingame.AbstractInventoryScreen;
+import net.minecraft.client.util.math.MatrixStack;
 import net.minecraft.command.CommandSource;
 import net.minecraft.datafixer.TypeReferences;
 import net.minecraft.entity.player.PlayerInventory;
 import net.minecraft.item.ItemStack;
 import net.minecraft.nbt.*;
+import net.minecraft.screen.PlayerScreenHandler;
 import net.minecraft.screen.slot.Slot;
+import net.minecraft.screen.slot.SlotActionType;
 import net.minecraft.server.command.ServerCommandSource;
+import net.minecraft.text.LiteralText;
 import net.minecraft.text.TranslatableText;
+import net.minecraft.util.Formatting;
 import net.minecraft.util.Util;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
@@ -82,7 +89,11 @@ public class KitCommand {
                                         .executes(ctx -> load(ctx.getSource(), StringArgumentType.getString(ctx, "name"), true)))
                                 .executes(ctx -> load(ctx.getSource(), StringArgumentType.getString(ctx, "name"), false))))
                 .then(literal("list")
-                        .executes(ctx -> list(ctx.getSource()))));
+                        .executes(ctx -> list(ctx.getSource())))
+                .then(literal("preview")
+                        .then(argument("name", StringArgumentType.string())
+                                .suggests((ctx, builder) -> CommandSource.suggestMatching(kits.keySet(), builder))
+                                .executes(ctx -> preview(ctx.getSource(), StringArgumentType.getString(ctx, "name"))))));
     }
 
     private static int create(ServerCommandSource source, String name) throws CommandSyntaxException {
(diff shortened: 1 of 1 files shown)


src/main/java/net/earthcomputer/clientcommands/command/KitCommand.java
```diff
@@ -192,3 +214,38 @@ private static void loadFile() throws IOException {
...
+
+    public PreviewScreen(PlayerScreenHandler playerScreenHandler, PlayerInventory inventory, String name) {
+        super(playerScreenHandler, inventory, new LiteralText("kit: " + name).styled(style -> style.withColor(Formatting.RED)));
+        this.passEvents = true;
+        this.titleX = 80;
+    }
+
+    protected void init() {
+        super.init();
+    }
```
Earthcomputer: Unnecessary method

src/main/java/net/earthcomputer/clientcommands/command/KitCommand.java
```diff
@@ -192,3 +214,38 @@ private static void loadFile() throws IOException {
...
+
+    protected void drawBackground(MatrixStack matrices, float delta, int mouseX, int mouseY) {
+        RenderSystem.color4f(1.0F, 1.0F, 1.0F, 1.0F);
+        this.client.getTextureManager().bindTexture(BACKGROUND_TEXTURE);
+        this.drawTexture(matrices, this.x, this.y, 0, 0, this.backgroundWidth, this.backgroundHeight);
+    }
+
+    public void removed() {
+        super.removed();
+    }
```
Earthcomputer: Also unnecessary

src/main/java/net/earthcomputer/clientcommands/command/KitCommand.java
```diff
@@ -192,3 +214,38 @@ private static void loadFile() throws IOException {
         }
     }
 }
+
+class PreviewScreen extends AbstractInventoryScreen<PlayerScreenHandler> {
+
+    public PreviewScreen(PlayerScreenHandler playerScreenHandler, PlayerInventory inventory, String name) {
+        super(playerScreenHandler, inventory, new LiteralText("kit: " + name).styled(style -> style.withColor(Formatting.RED)));
```
Earthcomputer: The screen title should probably be translatable

src/main/java/net/earthcomputer/clientcommands/command/KitCommand.java
```diff
@@ -192,3 +214,38 @@ private static void loadFile() throws IOException {
...
+        this.passEvents = true;
+        this.titleX = 80;
+    }
+
+    protected void init() {
+        super.init();
+    }
+
+    protected void drawForeground(MatrixStack matrices, int mouseX, int mouseY) {
+        this.textRenderer.draw(matrices, this.title, (float) this.titleX, (float) this.titleY, 4210752);
```
Earthcomputer: Colours should be in hex

src/main/java/net/earthcomputer/clientcommands/command/KitCommand.java
```diff
@@ -151,6 +161,18 @@ private static int list(ServerCommandSource source) {
...
 
+    private static int preview(ServerCommandSource source, String name) throws CommandSyntaxException {
+        ListTag kit = kits.get(name);
+        if (kit == null) {
+            throw NOT_FOUND_EXCEPTION.create(name);
+        }
+
+        PlayerInventory tempInv = new PlayerInventory(client.player);
+        tempInv.deserialize(kit);
+        client.send(() -> client.openScreen(new PreviewScreen(new PlayerScreenHandler(tempInv, true, client.player), tempInv, name)));
```
Earthcomputer, line 178: Is there any reason the opening of the screen has to be deferred? Why not just call `client.openScreen` here directly?
Also, is using a player screen handler likely to have any side effects?

xpple: It has to be. In your Discord I posted an explanation for why this has to be like this. I would link to it but I’m on my phone at the moment. Will get back to you

src/main/java/net/earthcomputer/clientcommands/command/KitCommand.java
xpple: I mean... the literal of the command is literally `ckit` 🤷‍♂️

src/main/java/net/earthcomputer/clientcommands/command/KitCommand.java
xpple: So, there reason is this:
After executing a command, the current screen will be closed (the chat hud). But if you open a new screen in a command, that new screen will be closed instantly along with the chat hud. To _fix_ this issue some people registered a certain event on the client's tick:
```java
ClientTickEvents.START_CLIENT_TICK.register(client -> {
    if (KitCommand.openScreen != null) {
        client.openScreen(KitCommand.openScreen);
        KitCommand.openScreen = null;
    }
});
```
I kind of hate this solution to be honest. Using `client.send()` is a bit cleaner, in my opinion.

In regards to instantiating a `new PlayerScreenHandler`, I don't know - sorry. It was either this or creating an entire `ClientPlayerEntity`, which is probably worse. Another option would be to create an entirely new `Screen`, but this suffices for now, I guess.

Earthcomputer: Who the heck suggested to use the tick event to fix that? Nope, you've done it the right way, nice one :+1: 
You should add a comment to explain why this workaround is needed though, so someone reading the code will know.

xpple: Everyone whom I asked suggested this 🤣. I refused to use that.
- [OroArmor](https://discord.com/channels/507304429255393322/507982478276034570/829625729213792286)
- [Random Forge people apparently](https://discord.com/channels/703363189081112647/703598704266969099/830223532109332570)

I'll add a comment.

Earthcomputer (approved):
Thank you!

PR #233, opened 2021-04-13
Add CONTRIBUTING

xpple:
In regards to
> And the UPPER_SNAKE_CASE (not :scream: :snake: case) is for all immutable static final fields

I don't think this applied to all immutable static final fields, like the `client` field in `KitCommand`. What should I do with it?


CONTRIBUTING.md
```diff
@@ -0,0 +1,55 @@
...
+## Scope
+clientcommands is not a hacked or cheat client. Commands or utilities whose sole purpose is to provide an unfair advantage over other players are out of scope for clientcommands.
+
+That said, commands which are only needed in very specific circumstances are also out of scope. clientcommands consists of commands that are generally useful. For edge cases there is the [scripting API](https://github.com/Earthcomputer/clientcommands/blob/fabric/docs/clientcommands.ts) contained within this mod.
+## Code style
+Please adhere to these styling instructions when contributing. If you don't do this initially you will be requested to change your commits accordingly.
+These instructions are not foundational. You're expected to have an understanding of what is clean Java code and what isn't. These instructions build on general consensus.
+### Variables
+#### Naming
+* Use descriptive variable names everywhere, except when they are meaningless (loop-indices etc.)
```
Earthcomputer: My rule of thumb for loop indices is the following:
- Never use `j`. If you find yourself wanting to use `j`, it's a sign that you should rename your loop variables to something more descriptive. You can use `i` but no more than that.
- Loops with small bodies (not more than a few lines) may use `i` as the loop variable, unless there is something obvious that's more descriptive.
- Loops with large bodies may only use `i` if the loop variable is unused within the loop body.

CONTRIBUTING.md
```diff
@@ -0,0 +1,55 @@
...
+#### Command exceptions
+* Exception variables should use SCREAMING_SNAKE_CASE
+* Mark exception variables as `private static final`
+* They should always be translated
+### Statements
+#### `if`, `for` and `while` statements
+* `if`, `for` and `while` statements should always use braces
+* There should be a space between the keyword and the statement-bracket
+#### `import` statements
+* `import static`s should normally use wildcard imports
```
Earthcomputer: I would say "always" rather than "normally"

CONTRIBUTING.md
```diff
@@ -0,0 +1,55 @@
...
+   * They should be placed after the other imports
+   * Don't use `import static` for anything else than commands
+* Use import statements rather than fully qualified class names
+### Command registry
+* Execute a command's code mostly in a dedicated method, rather than in the `register` method
+* If the command was successful and there isn't something sensible to return, it should `return 0`. If the command was not successful, an exception should be thrown
+### Miscellaneous
+* Only use up to Java 8 in your code
+* Java classes should have a newline at the end of the file
```
Earthcomputer: All files

CONTRIBUTING.md
```diff
@@ -0,0 +1,55 @@
...
+
+That said, commands which are only needed in very specific circumstances are also out of scope. clientcommands consists of commands that are generally useful. For edge cases there is the [scripting API](https://github.com/Earthcomputer/clientcommands/blob/fabric/docs/clientcommands.ts) contained within this mod.
+## Code style
+Please adhere to these styling instructions when contributing. If you don't do this initially you will be requested to change your commits accordingly.
+These instructions are not foundational. You're expected to have an understanding of what is clean Java code and what isn't. These instructions build on general consensus.
+### Variables
+#### Naming
+* Use descriptive variable names everywhere, except when they are meaningless (loop-indices etc.)
+#### Command exceptions
+* Exception variables should use SCREAMING_SNAKE_CASE
```
Earthcomputer: This should probably be generalized to all immutable constants. Also, all command related stuff should go in their own category. There's other stuff I'd like to put about the command style.
Also please call this UPPER_SNAKE_CASE

CONTRIBUTING.md
```diff
@@ -0,0 +1,55 @@
...
+* Mark exception variables as `private static final`
+* They should always be translated
```
Earthcomputer: This can be generalized too. Any message that the user may end up seeing needs to be translated. The only exception to this is if the text is used for debugging purposes, e.g. logging or info on the f3 screen etc.

Earthcomputer (commented):
There are definitely some other rules that I'd like to add, but I can accept this with the following changes for now, then I'll add them as I think of them.

CONTRIBUTING.md
xpple: Funny because I copied your exact words hahaha. Will change

CONTRIBUTING.md
```diff
@@ -0,0 +1,65 @@
...
+* `import static`s should always use wildcard imports
+   * They should be placed after the other imports
+   * Don't use `import static` for anything else than commands
+* Use import statements rather than fully qualified class names
+### Command registry
+#### General
+* Execute a command's code mostly in a dedicated method, rather than in the `register` method
+* If the command was successful and there isn't something sensible to return, it should `return 0`. If the command was not successful, an exception should be thrown
+* Any feedback the player may receive should be translated
+   * Except for debugging (log messages, debug HUD)
```
Earthcomputer: Maybe move this into miscellaneous? Also I'd use "messages" rather than "feedback".

Earthcomputer (approved):
Good enough for now, we can always improve this as we go

PR #236, opened 2021-04-17
Add the citemgroup command

### LINKED ISSUE #235: Custom creative tabs
So first, **I don't know if it's possible or not.** 
And second, yes, I can't think of a better name

### Usage 
```sh
/citem add item <item|slot> <id> <categoryName>
```
```sh
/citem remove item <id> <category>
```
`item` is mean the [item_stack](https://minecraft.fandom.com/wiki/Argument_types#minecraft:item_stack)
`category` is mean the category name, can be vanilla/default, or custom (see below)

```sh
/citem add category <categoryName> <displayName>  <icon|slot>
```
```sh
/citem remove category <categoryName>
```
`displayName` is mean the actual name for it, support [Raw JSON Format](https://minecraft.fandom.com/wiki/Raw_JSON_text_format)
`icon` is mean the icon for category
image
^ the icon is `lava_bucket`
`slot` is mean the [item_slot](https://minecraft.fandom.com/wiki/Argument_types#minecraft:item_slot)
```sh
/citem list <item|category>
```

#### Items
For adding it, run this command (for example)
`/citem add item debug_stick 1 misc`
It will add `debug_stick` to the `misc` category
we can also add `air`, we cannot take it, the air is just for organizing item
image
For example, the image above is `debug_stick`, `air`, `debug_stick`

Then, for removing it, run this command (for example)
`/citem remove item 1 misc`
It will delete the name of `1` (which in this case, is that debug stick) from the `misc` category. (And each category have a different name, so you can use the name `1` again in another category)

#### Category
For adding it, run this command (for example)
`/citem add category commandsOnly {"text":"Commands Only","bold":true} command_block{Enchantments:[{}]}`
It will add a category called **Commands Only** and the logo is command block (but shiny, because why not?)
And items can put into this category (eg. `/citem add debug_stick 2 commandsOnly`)

For removing it, run this command (for example)
`/citem remove category commandsOnly`
And basically, it will remove the categoryName of `commandsOnly` (which in this case, is that **Commands Only**)

#### List
For listing them, run this command (for example)
`/citem list item`
then, it will list out, all the item that we add

Oh also (I don't know if it's possible or not), can you add a search bar for each custom tabs?
Why?
Because for some reason custom added items to items tabs don't appear in regular search

xpple:
Closes #235 

This. Was. A. Lot. Of. Pain.

This took me absolutely ages to code, but I'm finally done. Dynamically changing item groups is HARD. I initally got this working with `ItemGroup#appendStacks()`, but that turned out to be abuse of the method. After that was no longer an option, I tried setting the `group` for an `Item`, which removed the item from the vanilla groups 🤣. The last option was to recreate the entire item group to "modify" it. While this may sound inefficient, I did manage to optimise it by only recreating the item groups >= the to be modified group.

### DIFF
--- a/src/main/java/net/earthcomputer/clientcommands/command/ItemGroupCommand.java
+++ b/src/main/java/net/earthcomputer/clientcommands/command/ItemGroupCommand.java
@@ -20,6 +20,7 @@ import net.minecraft.nbt.*;
 import net.minecraft.server.command.ServerCommandSource;
 import net.minecraft.text.TranslatableText;
 import net.minecraft.util.Identifier;
+import net.minecraft.util.InvalidIdentifierException;
 import net.minecraft.util.Util;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
@@ -45,6 +46,7 @@ public class ItemGroupCommand {
     private static final DynamicCommandExceptionType OUT_OF_BOUNDS_EXCEPTION = new DynamicCommandExceptionType(arg -> new TranslatableText("commands.citemgroup.outOfBounds", arg));
 
     private static final SimpleCommandExceptionType SAVE_FAILED_EXCEPTION = new SimpleCommandExceptionType(new TranslatableText("commands.citemgroup.saveFile.failed"));
+    private static final DynamicCommandExceptionType ILLEGAL_CHARACTER_EXCEPTION = new DynamicCommandExceptionType(arg -> new TranslatableText("commands.citemgroup.addGroup.illegalCharacter", arg));
     private static final DynamicCommandExceptionType ALREADY_EXISTS_EXCEPTION = new DynamicCommandExceptionType(arg -> new TranslatableText("commands.citemgroup.addGroup.alreadyExists", arg));
 
     private static final Path configPath = FabricLoader.getInstance().getConfigDir().resolve("clientcommands");
@@ -98,14 +100,18 @@ public class ItemGroupCommand {
             throw ALREADY_EXISTS_EXCEPTION.create(name);
         }
 
-        ItemGroup itemGroup = FabricItemGroupBuilder.create(
-                new Identifier("clientcommands", name))
-                .icon(() -> icon)
-                .build();
+        try {
+            ItemGroup itemGroup = FabricItemGroupBuilder.create(
+                    new Identifier("clientcommands", name))
+                    .icon(() -> icon)
+                    .build();
 
-        groups.put(name, new Group(itemGroup, icon, new ListTag()));
-        saveFile();
-        sendFeedback("commands.citemgroup.addGroup.success", name);
+            groups.put(name, new Group(itemGroup, icon, new ListTag()));
+            saveFile();
+            sendFeedback("commands.citemgroup.addGroup.success", name);
+        } catch (InvalidIdentifierException e) {
+            throw ILLEGAL_CHARACTER_EXCEPTION.create(name);
+        }
         return 0;
     }
 
--- a/src/main/java/net/earthcomputer/clientcommands/command/ItemGroupCommand.java
+++ b/src/main/java/net/earthcomputer/clientcommands/command/ItemGroupCommand.java
@@ -27,21 +27,15 @@ import org.apache.logging.log4j.Logger;
 import java.io.File;
 import java.io.IOException;
 import java.nio.file.Path;
-import java.util.SortedMap;
-import java.util.TreeMap;
+import java.util.HashMap;
+import java.util.Map;
 
-import static com.mojang.brigadier.arguments.IntegerArgumentType.getInteger;
-import static com.mojang.brigadier.arguments.IntegerArgumentType.integer;
-import static com.mojang.brigadier.arguments.StringArgumentType.getString;
-import static com.mojang.brigadier.arguments.StringArgumentType.string;
-import static net.earthcomputer.clientcommands.command.ClientCommandManager.addClientSideCommand;
-import static net.earthcomputer.clientcommands.command.ClientCommandManager.sendFeedback;
-import static net.minecraft.command.CommandSource.suggestMatching;
-import static net.minecraft.command.argument.ItemStackArgumentType.getItemStackArgument;
-import static net.minecraft.command.argument.ItemStackArgumentType.itemStack;
-import static net.minecraft.server.command.CommandManager.argument;
-import static net.minecraft.server.command.CommandManager.literal;
-
+import static com.mojang.brigadier.arguments.IntegerArgumentType.*;
+import static com.mojang.brigadier.arguments.StringArgumentType.*;
+import static net.earthcomputer.clientcommands.command.ClientCommandManager.*;
+import static net.minecraft.command.CommandSource.*;
+import static net.minecraft.command.argument.ItemStackArgumentType.*;
+import static net.minecraft.server.command.CommandManager.*;
 
 public class ItemGroupCommand {
 
@@ -57,7 +51,7 @@ public class ItemGroupCommand {
 
     private static final MinecraftClient client = MinecraftClient.getInstance();
 
-    private static final SortedMap<String, Group> groups = new TreeMap<>();
+    private static final Map<String, Group> groups = new HashMap<>();
 
     static {
         try {
--- a/src/main/java/net/earthcomputer/clientcommands/command/ItemGroupCommand.java
+++ b/src/main/java/net/earthcomputer/clientcommands/command/ItemGroupCommand.java
@@ -106,18 +106,17 @@ public class ItemGroupCommand {
             throw ALREADY_EXISTS_EXCEPTION.create(name);
         }
 
-        try {
-            ItemGroup itemGroup = FabricItemGroupBuilder.create(
-                    new Identifier("clientcommands", name))
-                    .icon(() -> icon)
-                    .build();
-
-            groups.put(name, new Group(itemGroup, icon, new ListTag()));
-            saveFile();
-            sendFeedback("commands.citemgroup.addGroup.success", name);
-        } catch (InvalidIdentifierException e) {
+        final Identifier identifier = Identifier.tryParse("clientcommands:" + name);
+        if (identifier == null) {
             throw ILLEGAL_CHARACTER_EXCEPTION.create(name);
         }
+        ItemGroup itemGroup = FabricItemGroupBuilder.create(identifier)
+                .icon(() -> icon)
+                .build();
+
+        groups.put(name, new Group(itemGroup, icon, new ListTag()));
+        saveFile();
+        sendFeedback("commands.citemgroup.addGroup.success", name);
         return 0;
     }
 
--- a/src/main/java/net/earthcomputer/clientcommands/command/ItemGroupCommand.java
+++ b/src/main/java/net/earthcomputer/clientcommands/command/ItemGroupCommand.java
@@ -84,7 +84,13 @@ public class ItemGroupCommand {
                             .then(argument("itemstack", itemStack())
                                 .then(argument("count", integer(1))
                                     .executes(ctx -> setStack(ctx.getSource(), getString(ctx, "group"), getInteger(ctx, "index"), getItemStackArgument(ctx, "itemstack").createStack(getInteger(ctx, "count"), false))))
-                                .executes(ctx -> setStack(ctx.getSource(), getString(ctx, "group"), getInteger(ctx, "index"), getItemStackArgument(ctx, "itemstack").createStack(1, false))))))))
+                                .executes(ctx -> setStack(ctx.getSource(), getString(ctx, "group"), getInteger(ctx, "index"), getItemStackArgument(ctx, "itemstack").createStack(1, false))))))
+                    .then(literal("icon")
+                        .then(argument("icon", itemStack())
+                            .executes(ctx -> changeIcon(ctx.getSource(), getString(ctx, "group"), getItemStackArgument(ctx, "icon").createStack(1, false)))))
+                    .then(literal("rename")
+                        .then(argument("new", string())
+                            .executes(ctx -> renameGroup(ctx.getSource(), getString(ctx, "group"), getString(ctx, "new")))))))
             .then(literal("add")
                 .then(argument("group", string())
                     .then(argument("icon", itemStack())
@@ -142,7 +148,7 @@ public class ItemGroupCommand {
 
         reloadGroups(itemGroup.getIndex());
         saveFile();
-        sendFeedback("commands.citemgroup.addStack.success", itemStack.getItem().getName().getString(), name);
+        sendFeedback("commands.citemgroup.addStack.success", itemStack.getItem().getName(), name);
         return 0;
     }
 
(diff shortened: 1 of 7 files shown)


src/main/java/net/earthcomputer/clientcommands/command/ItemGroupCommand.java
```diff
@@ -0,0 +1,307 @@
...
+
+    private static final MinecraftClient client = MinecraftClient.getInstance();
+
+    private static final Map<String, Group> groups = new HashMap<>();
+
+    static {
+        try {
+            loadFile();
+        } catch (IOException e) {
+            LOGGER.info("Could not load groups file, hence /citemgroup will not work!");
```
Earthcomputer:
```suggestion
            LOGGER.error("Could not load groups file, hence /citemgroup will not work!", e);
```

src/main/java/net/earthcomputer/clientcommands/command/ItemGroupCommand.java
```diff
@@ -0,0 +1,307 @@
...
+        }
+    }
+
+    public static void register(CommandDispatcher<ServerCommandSource> dispatcher) {
+        addClientSideCommand("citemgroup");
+
+        LiteralCommandNode<ServerCommandSource> citemgroup = dispatcher.register(literal("citemgroup"));
+        dispatcher.register(literal("citemgroup")
```
Earthcomputer, line 70: Can some of these indents below are twice as much as they should be, code shouldn't the this far right!

src/main/java/net/earthcomputer/clientcommands/mixin/CreativeInventoryScreenAccessor.java
```diff
@@ -0,0 +1,19 @@
...
+
+@Mixin(targets = "net.minecraft.client.gui.screen.ingame.CreativeInventoryScreen")
+public interface CreativeInventoryScreenAccessor {
+
+    @Accessor("selectedTab")
+    static void setSelectedTab(int selectedTab) {
+        throw new AssertionError();
+    }
+
+    @SuppressWarnings("AccessorTarget")
```
Earthcomputer, line 15: You can use `@Dynamic` here instead

src/main/java/net/earthcomputer/clientcommands/mixin/MixinItemGroup.java
```diff
@@ -0,0 +1,40 @@
...
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+
+@Mixin(ItemGroup.class)
+public abstract class MixinItemGroup implements IItemGroup {
+
+    @Shadow @Final @Mutable public static ItemGroup[] GROUPS;
+
+    @Mutable @Shadow @Final private Text translationKey;
```
Earthcomputer: The order of these annotations is inconsistent

src/main/java/net/earthcomputer/clientcommands/mixin/MixinItemGroup.java
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    @Override
+    public int getLength() {
+        return GROUPS.length;
+    }
+
+    @Inject(method = "<init>", at = @At("TAIL"))
+    private void substring(int index, String id, CallbackInfo ci) {
+        if (id.contains("clientcommands.")) {
```
Earthcomputer: Maybe `startsWith` instead?

Earthcomputer (changes requested):
I'm not sure about the whole item-groups-are-lists thing either. I'd prefer if they were sets you can add to, or remove from by an item predicate. Also, subcommands for renaming an item group, changing its icon, and duplicating one would be nice.

xpple:
The above commit also implements [this requested change](https://github.com/Earthcomputer/clientcommands/pull/236#discussion_r632073612).

xpple:
Working on this still
> Also, subcommands for renaming an item group, changing its icon, and duplicating one would be nice.

src/main/java/net/earthcomputer/clientcommands/command/ItemGroupCommand.java
```diff
@@ -0,0 +1,361 @@
...
+    }
+
+    private static int removeGroup(ServerCommandSource source, String name) throws CommandSyntaxException {
+        if (!groups.containsKey(name)) {
+            throw NOT_FOUND_EXCEPTION.create(name);
+        }
+
+        groups.remove(name);
+
+        reloadGroups(12);
```
Earthcomputer: Hardcoding this value 12 seems dangerous.What if other mods add item groups? What if a new Minecraft update adds new item groups? There must be a better way to do this, perhaps by filtering out the clientcommands groups?

src/main/java/net/earthcomputer/clientcommands/command/ItemGroupCommand.java
```diff
@@ -0,0 +1,361 @@
...
+    }
+
+    private static int setStack(ServerCommandSource source, String name, int index, ItemStack itemStack) throws CommandSyntaxException {
+        if (!groups.containsKey(name)) {
+            throw NOT_FOUND_EXCEPTION.create(name);
+        }
+
+        Group group = groups.get(name);
+        ListTag items = group.getItems();
+        if ((index < 0) || (index > items.size())) {
```
Earthcomputer:
```suggestion
        if ((index < 0) || (index >= items.size())) {
```

src/main/java/net/earthcomputer/clientcommands/command/ItemGroupCommand.java
```diff
@@ -0,0 +1,361 @@
...
+    }
+
+    private static int removeStack(ServerCommandSource source, String name, int index) throws CommandSyntaxException {
+        if (!groups.containsKey(name)) {
+            throw NOT_FOUND_EXCEPTION.create(name);
+        }
+
+        Group group = groups.get(name);
+        ListTag items = group.getItems();
+        if (index < 0 || index > items.size()) {
```
Earthcomputer:
```suggestion
        if (index < 0 || index >= items.size()) {
```

src/main/java/net/earthcomputer/clientcommands/command/ItemGroupCommand.java
```diff
@@ -0,0 +1,361 @@
...
+        ListTag items = group.getItems();
+        ItemStack icon = itemGroup.getIcon();
+        try {
+            itemGroup = FabricItemGroupBuilder.create(
+                    new Identifier("clientcommands", _new))
+                    .icon(() -> icon)
+                    .build();
+
+            groups.put(_new, new Group(itemGroup, icon, items));
+        } catch (InvalidIdentifierException e) {
```
Earthcomputer, line 227: Maybe `Identifier.tryParse` could be used here instead of a try-catch?

src/main/java/net/earthcomputer/clientcommands/command/ItemGroupCommand.java
```diff
@@ -0,0 +1,361 @@
...
+            });
+            rootTag.putInt("DataVersion", SharedConstants.getGameVersion().getWorldVersion());
+            rootTag.put("Groups", compoundTag);
+            File newFile = File.createTempFile("groups", ".dat", configPath.toFile());
+            NbtIo.write(rootTag, newFile);
+            File backupFile = new File(configPath.toFile(), "groups.dat_old");
+            File currentFile = new File(configPath.toFile(), "groups.dat");
+            Util.backupAndReplace(currentFile, newFile, backupFile);
+        } catch (IOException e) {
+            throw SAVE_FAILED_EXCEPTION.create();
```
Earthcomputer, line 256: Printing the stack trace here might be useful to help people figure out what's going wrong

src/main/java/net/earthcomputer/clientcommands/mixin/MixinItemGroup.java
```diff
@@ -0,0 +1,40 @@
...
+
+    @Shadow @Mutable @Final public static ItemGroup[] GROUPS;
+
+    @Shadow @Mutable @Final private Text translationKey;
+
+    @Override
+    public void shrink(int index) {
+        ItemGroup[] tempGroups = GROUPS;
+        GROUPS = new ItemGroup[index];
+        System.arraycopy(tempGroups, 0, GROUPS, 0, GROUPS.length);
```
Earthcomputer: This can be simplified to `Arrays.copyOf`. Also, was there a reason that `GROUPS` couldn't be accessed by an accessor mixin instead?

src/main/java/net/earthcomputer/clientcommands/command/ItemGroupCommand.java
```diff
@@ -0,0 +1,361 @@
...
+        }
+
+        Group group = groups.get(name);
+        ListTag items = group.getItems();
+        ItemGroup itemGroup = group.getItemGroup();
+        items.add(itemStack.toTag(new CompoundTag()));
+
+        reloadGroups(itemGroup.getIndex());
+        saveFile();
+        sendFeedback("commands.citemgroup.addStack.success", itemStack.getItem().getName(), name);
```
Earthcomputer, line 149: Is `getName()` the same method that's used to display weapons in death messages?

Earthcomputer (changes requested):

src/main/java/net/earthcomputer/clientcommands/command/ItemGroupCommand.java
xpple: That returns `null` if it fails, which is not what we want here, do we?

src/main/java/net/earthcomputer/clientcommands/mixin/MixinItemGroup.java
xpple: `Arrays.copyOf` calls `System.arraycopy` internally, are you sure you want to use it?

As to the other question, I guess it could be 🤷‍♂️.

src/main/java/net/earthcomputer/clientcommands/command/ItemGroupCommand.java
xpple: Ohh yeah maybe it will return the name of the item if it has been renamed. Let me check that.

xpple: Nope, it won't. This will always return the default name of the item. See the below image for reference:
image

src/main/java/net/earthcomputer/clientcommands/command/ItemGroupCommand.java
xpple: Yeah this is difficult. If other mods would add an item group, it would be deleted. I will try to implement a check for clientcommands-created item groups. I do not worry as much about a Minecraft update adding a new group, that will just change the starting index (12 now).

xpple: Does this look reasonable? Edit: the below code is not functional.
```java
@Override
public void shrink(int index) {
    ItemGroup[] tempGroups = GROUPS;
    tempGroups = Arrays.stream(tempGroups).filter(itemGroup -> !itemGroup.getName().startsWith("clientcommands.")).toArray(ItemGroup[]::new);
    GROUPS = new ItemGroup[index + tempGroups.length];
    System.arraycopy(tempGroups, 0, GROUPS, 0, GROUPS.length);
}
```

xpple: To be honest I'm going quite insane from this mess of arrays and indices.

xpple:
I've implemented most of the requested change, only not those who are still to be discussed. Anyway, to solve the issue proposed [here](https://github.com/Earthcomputer/clientcommands/pull/236#discussion_r637886623), I had to get rid of the initial optimisation I implemented. It's either the optimisation or compatibility with other mods that add item groups.

src/main/java/net/earthcomputer/clientcommands/command/ItemGroupCommand.java
xpple: Update to clarify: the most recent commit fixed this issue.

src/main/java/net/earthcomputer/clientcommands/command/ItemGroupCommand.java
Earthcomputer: Idk I find it cleaner

src/main/java/net/earthcomputer/clientcommands/mixin/MixinItemGroup.java
```diff
@@ -0,0 +1,43 @@
...
+public abstract class MixinItemGroup implements IItemGroup {
+
+    @Shadow @Mutable @Final public static ItemGroup[] GROUPS;
+
+    @Shadow @Mutable @Final private Text translationKey;
+
+    @Override
+    public void shrink() {
+        ItemGroup[] tempGroups = GROUPS;
+        tempGroups = Arrays.stream(tempGroups).filter(itemGroup -> !itemGroup.getName().startsWith("clientcommands.")).toArray(ItemGroup[]::new);
```
Earthcomputer: `toArray(Supplier)` is not a Java 8 method

src/main/java/net/earthcomputer/clientcommands/mixin/MixinItemGroup.java
```diff
@@ -0,0 +1,43 @@
...
+        GROUPS = new ItemGroup[tempGroups.length];
+        System.arraycopy(tempGroups, 0, GROUPS, 0, GROUPS.length);
```
Earthcomputer: Why not just assign to `GROUPS` directly rather than using a temp array?

src/main/java/net/earthcomputer/clientcommands/interfaces/IItemGroup.java
```diff
@@ -0,0 +1,7 @@
+package net.earthcomputer.clientcommands.interfaces;
+
+public interface IItemGroup {
+    int getLength();
```
Earthcomputer: `getLength` is never used

src/main/java/net/earthcomputer/clientcommands/mixin/CreativeInventoryScreenAccessor.java
Earthcomputer: Why was this resolved?

src/main/java/net/earthcomputer/clientcommands/mixin/MixinItemGroup.java
xpple: Yes it is [(see the API documentation)](https://docs.oracle.com/javase/8/docs/api/java/util/stream/Stream.html#toArray-java.util.function.IntFunction-).

src/main/java/net/earthcomputer/clientcommands/mixin/MixinItemGroup.java
xpple: Like this:
```java
GROUPS = Arrays.stream(GROUPS).filter(itemGroup -> !itemGroup.getName().startsWith("clientcommands.")).toArray(ItemGroup[]::new);
```
?

src/main/java/net/earthcomputer/clientcommands/mixin/MixinItemGroup.java
xpple: I think I missed your response here, what was it again?

Earthcomputer: The way it's been rewritten, you can just reassign `GROUPS` like I commented below.

src/main/java/net/earthcomputer/clientcommands/mixin/MixinItemGroup.java
Earthcomputer: Yep

src/main/java/net/earthcomputer/clientcommands/mixin/MixinItemGroup.java
Earthcomputer: Oh right, I forgot this was a stream rather than a collection

src/main/java/net/earthcomputer/clientcommands/mixin/MixinItemGroup.java
xpple: Kinda stupid I missed that 🤣, problems with modifying existing code I guess.

src/main/java/net/earthcomputer/clientcommands/mixin/CreativeInventoryScreenAccessor.java
xpple: Because I added the `@Dynamic` annotation. For your information, if I remember correctly, the `@Dynamic` annotation can't be used _instead of_ the `@SuppressWarnings` annotation: it's complementairy.

Earthcomputer (approved):
Lgtm

PR #238, opened 2021-04-26
Add the careastats command

### LINKED ISSUE #200: Suggsetion: counting blocks with countblocks command
Command /ccountblocks \<block id> \<pos1> \<pos2> that counts how many of that block is in that area, and maybe /ccountentity or /ccount \<entity> that does the same things as /cfind but without filling the chat with all entity positions.

For counting blocks it might check if that area is rendered or not and only counts the blocks in a fully rendered area.

xpple:
Closes #200 

After a very long process of getting things to work, I'm finally done. A few things first.
1. Yes - the code looks concerning. I did this on purpose though, to make it as efficient as possible (only getting every chunk once, making nearly everything `final`, etc.). For big areas, this command can take up to 1 second to complete, which is a lot. Efficiency is key here. The time could be reduced by replacing all the `loop()` calls with a loop (about .1 second faster), but that _really_ makes things awful to look at.
2. I decided to make my own `SimpleBlockPredicateArgumentType` mostly because this command is not meant to check for any custom data. If it would, it would slow down the progress drastically.
3. I also added (with the help of CreepyCre) a new `ArgumentType`: `ListArgumentType`. This addition is really nice and powerful and could be used a lot in the future. For any future users of this `ArgumentType`: keep in mind that `getList()` returns a list of whatever the `parse()` method returns, **not** what the `get<...>` method returns


xpple:
Fuck I'm stupid, this relies on my previous PR

xpple:
@Earthcomputer if you are planning to merge this, please look at [this PR on my PR](https://github.com/xpple/clientcommands/pull/1/) first. If you like the idea of having more type safe arguments (for the ListArgumentType), I will merge it in to this PR.

Earthcomputer:
@xpple I think a better approach than that PR you linked, would be to pass a `Class<U>` into `getList`

xpple:
Sure, however that doesn't reflect Minecraft's style of creating arguments. (Which isn't really a problem to me but I do recall you saying you want to reflect that style.)

Earthcomputer:
It's more an extension of their style, than something which doesn't match. There is sometimes a balance in programming between conflicting code styles or requirements, where it's impossible to satisfy all of them perfectly. In those cases you have to use your best judgement to know to what extent it is worth following each style.

xpple:
Do you want me to implement that, or keep it how it is right now?

Earthcomputer:
Honestly it's fine as it is right now, a Class parameter might get annoying since you can't do `Predicate<Block>.class`

xpple:
Then this should be ready to be merged right?

Earthcomputer:
You said to look at that other PR before looking at this one, so I hadn't looked at this one yet. I'll take a look later when I have more time.

xpple:
Ohhh like that, alright. I'll resolve the conflicts already.

src/main/resources/assets/clientcommands/lang/en_us.json
```diff
@@ -1,4 +1,9 @@
 {
+  "commands.careastats.notLoaded": "A chunk in this area is not loaded",
+  "commands.careastats.output.1": "Scanned %d chunks in %dns ≈ %dms",
+  "commands.careastats.output.2": "Matched %d out of %d total blocks",
+  "commands.careastats.output.3": "Found %d entities in this area",
```
Earthcomputer: These could have more descriptive unlocalized names

src/main/java/net/earthcomputer/clientcommands/command/arguments/SimpleBlockPredicateArgumentType.java
```diff
@@ -0,0 +1,72 @@
...
+import net.minecraft.server.command.ServerCommandSource;
+import net.minecraft.tag.BlockTags;
+import net.minecraft.util.registry.Registry;
+
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.concurrent.CompletableFuture;
+import java.util.function.Predicate;
+
+public class SimpleBlockPredicateArgumentType implements ArgumentType<Predicate<Block>> {
```
Earthcomputer: Clientcommands already has `ClientBlockPredicateArgumentType`

src/main/java/net/earthcomputer/clientcommands/command/AreaStatsCommand.java
```diff
@@ -0,0 +1,205 @@
...
+            for (int chunkX = minMinChunk.getPos().x + 1; chunkX < maxMinChunk.getPos().x; chunkX++) {
+                for (int chunkZ = minMinChunk.getPos().z + 1; chunkZ < minMaxChunk.getPos().z; chunkZ++) {
+                    chunks++;
+                    final WorldChunk chunk = world.getChunk(chunkX, chunkZ);
+                    blocks = loop(16 * chunkX, 16 * chunkX + 15, 16 * chunkZ, 16 * chunkZ + 15, minY, maxY, blocks, blockPredicate, chunk, mutablePos);
+                }
+            }
+        }
+
+        final long entities = StreamSupport.stream(world.getEntities().spliterator(), false)
```
Earthcomputer: Guava has a nice method:
```suggestion
        final long entities = Streams.stream(world.getEntities())
```

src/main/java/net/earthcomputer/clientcommands/command/AreaStatsCommand.java
```diff
@@ -0,0 +1,205 @@
...
+                    }
+                }
+            }
+        }
+        return blocks;
+    }
+
+    private static Predicate<Block> or(List<Predicate<Block>> predicates) {
+        return predicates.stream().reduce(t -> false, Predicate::or);
+    }
```
Earthcomputer: Guava has another useful method: `Predicates.or`

src/main/java/net/earthcomputer/clientcommands/command/AreaStatsCommand.java
```diff
@@ -0,0 +1,205 @@
...
+    private static int areaStats(ServerCommandSource source, BlockPos pos1, BlockPos pos2, Predicate<Block> blockPredicate) throws CommandSyntaxException {
+        return areaStats(source, pos1, pos2, Collections.singletonList(blockPredicate));
+    }
+
+    private static int areaStats(ServerCommandSource source, BlockPos pos1, BlockPos pos2, List<Predicate<Block>> blockPredicates) throws CommandSyntaxException {
+        final ClientWorld world = MinecraftClient.getInstance().world;
+        final ChunkManager chunkManager = world.getChunkManager();
+        if (!chunkManager.isChunkLoaded(pos1.getX() >> 4, pos1.getZ() >> 4) || !chunkManager.isChunkLoaded(pos2.getX() >> 4, pos2.getZ() >> 4)) {
+            throw NOT_LOADED_EXCEPTION.create();
+        }
```
Earthcomputer: This method assumes that the loaded area is a square

src/main/java/net/earthcomputer/clientcommands/command/AreaStatsCommand.java
```diff
@@ -0,0 +1,205 @@
...
+
+        final Predicate<Block> blockPredicate = or(blockPredicates);
+
+        final long startTime = System.nanoTime();
+
+        final WorldChunk chunk1 = world.getWorldChunk(pos1);
+        final WorldChunk chunk2 = world.getWorldChunk(pos2);
+
+        final int minX, maxX, minZ, maxZ, minY, maxY;
+        if (pos1.getX() <= pos2.getX()) {
```
Earthcomputer: This section of code can be rewritten to use `Math.min` and `Math.max`

src/main/java/net/earthcomputer/clientcommands/command/AreaStatsCommand.java
```diff
@@ -0,0 +1,205 @@
...
+            chunks++;
+            blocks = loop(minX, maxX, minZ, maxZ, minY, maxY, blocks, blockPredicate, chunk1, mutablePos);
+        } else if (chunk1.getPos().x == chunk2.getPos().x) {
+            chunks++;
+            final WorldChunk minMinChunk = world.getChunk(minX >> 4, minZ >> 4);
+            blocks = loop(minX, maxX, minZ, (minZ >> 4) * 16 + 15, minY, maxY, blocks, blockPredicate, minMinChunk, mutablePos);
+
+            chunks++;
+            final WorldChunk minMaxChunk = world.getChunk(minX >> 4, maxZ >> 4);
+            blocks = loop(minX, maxX, (maxZ >> 4) * 16, maxZ, minY, maxY, blocks, blockPredicate, minMaxChunk, mutablePos);
```
Earthcomputer: `loop` could be changed to return the number of blocks in that chunk, then you can use `blocks +=` here.

Earthcomputer (changes requested):

src/main/java/net/earthcomputer/clientcommands/command/arguments/SimpleBlockPredicateArgumentType.java
xpple: The argument I implemented has been simplified (hence `Simple`) to match `Block`s and effectively ignore the block's `Property`s and NBT data. These simplifications make the process quicker, which is really important for this command.

src/main/java/net/earthcomputer/clientcommands/command/AreaStatsCommand.java
xpple: Oh, nice. I did notice it was annotated as `@Beta`, though. Fine by me.

src/main/java/net/earthcomputer/clientcommands/command/AreaStatsCommand.java
xpple: I'm unsure how I'm supposed to use that 🤐.

src/main/java/net/earthcomputer/clientcommands/command/AreaStatsCommand.java
xpple: Good point. I'll check each chunk individually I guess.

src/main/java/net/earthcomputer/clientcommands/command/AreaStatsCommand.java
wagyourtail: > Returns a predicate that evaluates to true if any one of its components evaluates to true. The components are evaluated in order, and evaluation will be "short-circuited" as soon as a true predicate is found. It defensively copies the array passed in, so future changes to it won't alter the behavior of this predicate. If components is empty, the returned predicate will always evaluate to false.

https://guava.dev/releases/19.0/api/docs/com/google/common/base/Predicates.html#or(com.google.common.base.Predicate...)

it basically just chains `Predicate.or` https://docs.oracle.com/javase/8/docs/api/java/util/function/Predicate.html#or-java.util.function.Predicate-

Earthcomputer: Basically it does exactly what the function you've written does :P

xpple: I don't see the point of that at all. Plus IntelliJ says:
`Guava's functional primitives can be replaced with Java API`.

Earthcomputer: It means you don't need to write this function yourself. What does intellij suggest you replace guava's method with?

xpple: Do you find this cleaner?
```java
final Predicate<Block> blockPredicate = Predicates.or(blockPredicates.toArray(new com.google.common.base.Predicate[0]))::apply;
```
Notice the fully qualified class, and notice how IntelliJ is screaming at me 🤣.

src/main/java/net/earthcomputer/clientcommands/command/arguments/SimpleBlockPredicateArgumentType.java
Earthcomputer: Hmm, is there no way to optimize the existing argument type?

xpple: Well, it's rather annoying. I've seen the changes made to the argument and while that does allow for simpler checks, it still has some of the annoyances Minecraft's `BlockPredicateArgumentType` has. For example, it returns a `BlockPredicate` in the `parse()` method, and sadly not a `Predicate<Block>`. Also, Minecraft seems to use `CachedBlockPosition` and while that may seem nice, it's actually rediculously slow. Things like this make it annoying to work with.

src/main/java/net/earthcomputer/clientcommands/command/AreaStatsCommand.java
xpple: Resolving this as I couldn't find a way to use Guava's method.

src/main/java/net/earthcomputer/clientcommands/command/arguments/SimpleBlockPredicateArgumentType.java
xpple: I'm considering replacing `ClientBlockPredicateArgumentType` with `SimpleBlockPredicateArgumentType` and adding the optional `allowTags` and `allowNbt` to it, should I?

xpple: As discussed on Discord, adding optional nbt support turns out to be hard with the current setup. Tags are properly supported now (they weren't before), though. `BlockState`s are still ignored.

Earthcomputer (approved):
Awesome!

xpple:
Glad that's done!

PR #242, opened 2021-04-30
Add FormattedTextArgumentType

### LINKED ISSUE #208: [Feature] /cnote support Raw JSON Text Format
So now the /cnote support Raw JSON Text Format, if the player just type in like `/cnote test test` it will work without `"`. So it will function like the current one, but it support Raw JSON Text Format 

Also if you can, support markdown, because if you want to type like
test **bold**
you would have to type
`[{"text":"test"},{"text":"bold","bold":true}]`
Its kinda annoying
But if in markdown you just type
`test **bold**`

xpple:
Closes #208 

An argument type to easily type formatted text in the chat. Includes all types of formatting.
The proposed format by @barraIhsan (`.md`) didn't seem like a good option to me, especially because it naturally doesn't support any colour syntax.

### DIFF
--- a/src/main/java/net/earthcomputer/clientcommands/command/arguments/FormattedTextArgumentType.java
+++ b/src/main/java/net/earthcomputer/clientcommands/command/arguments/FormattedTextArgumentType.java
@@ -1,7 +1,6 @@
 package net.earthcomputer.clientcommands.command.arguments;
 
 import com.google.common.collect.ImmutableMap;
-import com.mojang.brigadier.LiteralMessage;
 import com.mojang.brigadier.StringReader;
 import com.mojang.brigadier.arguments.ArgumentType;
 import com.mojang.brigadier.context.CommandContext;
@@ -22,9 +21,9 @@ import java.util.function.Consumer;
 
 public class FormattedTextArgumentType implements ArgumentType<LiteralText> {
 
-    private static final Collection<String> EXAMPLES = Arrays.asList("Earth", "bold(xpple)", "bold(italic(red(nwex)))");
+    private static final Collection<String> EXAMPLES = Arrays.asList("Earth", "bold{xpple}", "bold{italic{red{nwex}}}");
 
-    private static final DynamicCommandExceptionType INVALID_ARGUMENT_EXCEPTION = new DynamicCommandExceptionType(arg -> new LiteralMessage("Invalid argument '" + arg + "'"));
+    private static final DynamicCommandExceptionType INVALID_ARGUMENT_EXCEPTION = new DynamicCommandExceptionType(arg -> new TranslatableText("commands.client.invalidArgument", arg));
 
     private FormattedTextArgumentType() {
     }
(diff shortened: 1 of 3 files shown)


xpple:
Would you want me to add aliases? Such as `strikethrough -> strike` or `obfuscated -> magic`.

xpple:
And what do you prefer, the current code (a big `Map`) or this (see under)?
```java
private enum FormattedText {

    AQUA("aqua", (s, o) -> s.withFormatting(Formatting.AQUA), 0);
    // ...

    private final String name;
    private final BiFunction<Style, List<String>, Style> operator;
    private final int argumentCount;
    private final String[] suggestions;

    private FormattedText(String name, BiFunction<Style, List<String>, Style> operator, int argumentCount) {
        this.operator = operator;
        this.argumentCount = argumentCount;
        this.suggestions = new String[]{};
    }

    private FormattedText(String name, BiFunction<Style, List<String>, Style> operator, int argumentCount, String[] suggestions) {
        this.operator = operator;
        this.argumentCount = argumentCount;
        this.suggestions = suggestions;
    }

    public static Optional<FormattedText> get(String name) {
        return Arrays.stream(values())
            .filter(styling-> styling.name == name)
            .findAny();
    }

    public static boolean has(String name) {
        return Arrays.stream(values())
            .filter(styling-> styling.name == name)
            .findAny()
            .isPresent();
    }
}
```

src/main/java/net/earthcomputer/clientcommands/command/arguments/FormattedTextArgumentType.java
```diff
@@ -0,0 +1,226 @@
...
+import net.minecraft.util.Identifier;
+
+import java.util.*;
+import java.util.concurrent.CompletableFuture;
+import java.util.function.BiFunction;
+import java.util.function.Consumer;
+
+public class FormattedTextArgumentType implements ArgumentType<LiteralText> {
+
+    private static final Collection<String> EXAMPLES = Arrays.asList("Earth", "bold(xpple)", "bold(italic(red(nwex)))");
```
Earthcomputer: These examples need updating to use braces instead of parentheses

src/main/java/net/earthcomputer/clientcommands/command/arguments/FormattedTextArgumentType.java
```diff
@@ -0,0 +1,226 @@
...
+
+    private static final DynamicCommandExceptionType INVALID_ARGUMENT_EXCEPTION = new DynamicCommandExceptionType(arg -> new LiteralMessage("Invalid argument '" + arg + "'"));
```
Earthcomputer: This needs i18n

src/main/java/net/earthcomputer/clientcommands/command/arguments/FormattedTextArgumentType.java
```diff
@@ -0,0 +1,226 @@
...
+        public FormattedText(BiFunction<Style, List<String>, Style> styler, LiteralText argument, List<String> optional) {
+            this.styler = styler;
+            this.argument = argument;
+            this.optional = optional;
+        }
+
+        public LiteralText style() throws CommandSyntaxException {
+            try {
+                return (LiteralText) this.argument.setStyle(this.styler.apply(this.argument.getStyle(), this.optional));
+            } catch (Exception e) {
```
Earthcomputer: What Exception are you expecting here?

src/main/java/net/earthcomputer/clientcommands/command/arguments/FormattedTextArgumentType.java
```diff
@@ -0,0 +1,226 @@
...
+                this.operator = operator;
+                this.argumentCount = argumentCount;
+                this.suggestions = new String[]{};
+            }
+
+            private Styler(BiFunction<Style, List<String>, Style> operator, int argumentCount, String[] suggestions) {
+                this.operator = operator;
+                this.argumentCount = argumentCount;
+                this.suggestions = suggestions;
+            }
```
Earthcomputer: Might be cleaner to use varargs rather than these constructor overloads

Earthcomputer (changes requested):

xpple:
I will implement these changes when I get the time. Could you give your opinion on [this optional change](https://github.com/Earthcomputer/clientcommands/pull/242#issuecomment-830173468) though?

Earthcomputer:
The map is better I think

xpple:
> Would you want me to add aliases? Such as `strikethrough -> strike` or `obfuscated -> magic`.

And could you answer this as well?

Earthcomputer:
I don't mind having aliases, although I've never heard `obfuscated` called `magic`

xpple:
That must be a Bukkit thing to call them that I guess 😄. [This website](http://ess.khhq.net/mc/) seems to use it as well.

src/main/java/net/earthcomputer/clientcommands/command/arguments/FormattedTextArgumentType.java
xpple: Ah, forgot to remove that when I changed the method of applying styling.

Earthcomputer (approved):
Lgtm

PR #243, opened 2021-04-30
Fix autocomplete for entities

### LINKED ISSUE #241: [Suggestion] UUID Autocomplete when looking at entity
So, in vanilla when you look at an entity, then there is their UUID in the autocomplete (for example /data get entity)
image
But, in client commands, it doesn't happen (for example /cgetdata entity )
image
Same position, same entity.

Yes, it already [said](https://discord.com/channels/703363189081112647/703363189081112650/809386447282241576) that it gonna add into client command, but its forgotten

xpple:
Closes #241


Earthcomputer (approved):
Ty

PR #244, opened 2021-04-30
Fix border outlines

### LINKED ISSUE #221: `/cglow area` highlighted box is wrong
The highlighted area of blocks is wrong!
for the `<from>` coordinates is off by `+Y+Z`
for the `<to>` coordinates its off by `+X`

xpple:
Closes #221

### DIFF
--- a/src/main/java/net/earthcomputer/clientcommands/command/GlowCommand.java
+++ b/src/main/java/net/earthcomputer/clientcommands/command/GlowCommand.java
@@ -134,7 +134,7 @@ public class GlowCommand {
                 boundingBoxes.replaceAll((box) -> box.offset(pos1));
             }
         } else {
-            boundingBoxes.add(new Box(pos1, pos2));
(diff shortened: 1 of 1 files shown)


Earthcomputer (commented):
Hmm, is there a better way of doing this, maybe with `Math.min` and `Math.max`?

xpple:
Yes, but it's slightly slower since the amount of if-statements doubles.

Earthcomputer:
What? No it doesn't

xpple:
`Math.min`/`Math.max` is just an ternary if-statement, which would have to be called 6 times. The current code only uses three if statements by making use of the `else` clause.
Also, I don’t mind to change, we’re talking nanoseconds here.

Earthcomputer:
The code is clearer though which is what really matters here. Also the JIT exists so it might not be 6 branches in practice

Earthcomputer (approved):
Ty!

PR #246, opened 2021-05-01
chelp paging implementation

L0um15:
Since clientcommands is growing fast in features i thought that implementation of paginator would be really nice idea.
By default i set up to 9 entries per page + page number information. Minecraft allows to display 10 items without opening chatbox.
Also i removed command arguments from paginator which are still accessible within `/chelp <commandname>`

Preview:

### DIFF
--- a/src/main/java/net/earthcomputer/clientcommands/command/CHelpCommand.java
+++ b/src/main/java/net/earthcomputer/clientcommands/command/CHelpCommand.java
@@ -7,6 +7,8 @@ import com.mojang.brigadier.exceptions.SimpleCommandExceptionType;
 import com.mojang.brigadier.tree.CommandNode;
 import net.earthcomputer.clientcommands.Page;
 import net.earthcomputer.clientcommands.Paginator;
+import net.minecraft.client.MinecraftClient;
+import net.minecraft.client.MinecraftClientGame;
 import net.minecraft.server.command.ServerCommandSource;
 import net.minecraft.text.LiteralText;
 import net.minecraft.text.TranslatableText;
@@ -42,7 +44,9 @@ public class CHelpCommand {
 
                 }
 
-                Paginator<String> paginator = new Paginator<String>(commandNames, 19);
+                MinecraftClient.getInstance().inGameHud.getChatHud().clear(false);
+
+                Paginator<String> paginator = new Paginator<String>(commandNames, calculatePageSize());
                 Page<String> page = paginator.getPage(1);
 
                 for (int i = 0; i < page.items.size(); i++) {
@@ -75,7 +79,9 @@ public class CHelpCommand {
                             }
                         }
 
-                        Paginator<String> paginator = new Paginator<String>(commandNames, 19);
+                        MinecraftClient.getInstance().inGameHud.getChatHud().clear(false);
+
+                        Paginator<String> paginator = new Paginator<String>(commandNames, calculatePageSize());
 
                         if(!paginator.isValidPage(userPage)) {
                             sendFeedback(new TranslatableText("commands.chelp.paging.incorrect", userPage));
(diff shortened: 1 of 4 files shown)


src/main/resources/assets/clientcommands/lang/en_us.json
```diff
@@ -151,6 +151,9 @@
   "enchCrack.insn.itemThrows.noDummy": "No dummy enchantment",
   "enchCrack.insn.enchantments": "Enchantments on item:",
 
+  "paginator.display.page": "Page number: %d of %d",
+  "paginator.incorrect.page": "Specified page is out of bounds",
```
haykam821: These should be grouped and named like `chelp` translation keys.

src/main/java/net/earthcomputer/clientcommands/command/CHelpCommand.java
```diff
@@ -25,39 +29,92 @@ public static void register(CommandDispatcher<ServerCommandSource> dispatcher) {
...
+                            sendFeedback(new TranslatableText("paginator.incorrect.page"));
+                            return 0;
+                        }
+                        Page<String> page = paginator.getPage(userPage);
+                        for (int i = 0; i < page.Items.size(); i++)
+                        {
+                            sendFeedback(page.Items.get(i));
+                        }
+
+                        sendFeedback(new TranslatableText("paginator.display.page", page.PageNumber, paginator.getPageCount()));
```
haykam821: I think it would be better to include clickable navigation text on the page display.

haykam821 (changes requested):

src/main/resources/assets/clientcommands/lang/en_us.json
L0um15: Paginator was written not only for `/chelp` but also for other possible use cases.

src/main/java/net/earthcomputer/clientcommands/command/CHelpCommand.java
L0um15: Ok.

L0um15:
Since paginator is now clickable, i increased maximum amount of items per page by 10 in order to fill minecraft chat.
Now by default: 19 items per page + page navigation.
obraz

src/main/resources/assets/clientcommands/lang/en_us.json
```diff
@@ -76,6 +76,11 @@
   "commands.cglow.entity.success": "Glowing %d entities",
   "commands.cglow.area.success": "Glowing %d area(s)",
 
+  "commands.chelp.paging.left": "<<",
+  "commands.chelp.paging.right": ">>",
+  "commands.chelp.paging.body": " | %d/%d | ",
+  "commands.chelp.paging.incorrect": "Page: %d does not exist",
```
haykam821:
```suggestion
  "commands.chelp.paging.incorrect": "Page %d does not exist",
```

haykam821 (changes requested):

Earthcomputer (changes requested):
I just checked this out, and I'd like a couple of changes before I dig into the code.
- On your gui scale, 10 lines might look good, but on mine it was too much. It surely must be possible to detect how many lines the chat can show and adjust how many lines are printed based on that.
- For a help command, you have to be kinda helpful assuming the user knows nothing. Could you add support for page headers, then use it to print some basic instruction on how to get help for each command?

Other than that, I've had a very brief look at the code. Could you please make sure you are following the code style of the rest of the project, including style on variable naming and brace placement?

L0um15:
> Could you please make sure you are following the code style of the rest of the project, including style on variable naming and brace placement?

Too much C# lately. I forgot that on java every thing should start lowercased with different brace placement than C#

> On your gui scale, 10 lines might look good, but on mine it was too much.

I should've expected that because i hardcoded value for maximum page entries. I will look into it.

> Could you add support for page headers, then use it to print some basic instruction on how to get help for each command

Yes, sure.

Earthcomputer:
You should probably be aware that this change is unlikely to persist once we move to using Fabric's client command API.

L0um15:
Paginator clears chat now without touching message history.

I added first page header with short information about usage.


Paginator is now flexible, it will scale alongside with GUI


It will also respect focused height of the chatbox

L0um15:
Since ClientCommands is planning to use Fabric Command API there is no need for paging implementation.

PR #260, opened 2021-05-16
Use the Fabric Command API to register client commands

haykam821:


Earthcomputer (approved):
LGTM. It's a shame that FabricClientCommandSource.getMeta seems to be useless. Btw what's fcapi?

PR #265, opened 2021-05-20
/ckit preview fix

### LINKED ISSUE #245: You can move item while in `ckit preview`
So, when you run `ckit preview <name>`, then it will open up the preview inventory GUI.
But if you have something in your inventory, you can move it while in preview mode.
If it's WAI, we should not move the item in preview mode.

https://user-images.githubusercontent.com/57800056/116779591-666fed80-aaa1-11eb-8b27-d18f48c2a5a6.mp4

xpple:
Closes #245


Earthcomputer:
Can you still throw items out by hovering over them and pressing Q? What about using the number keys to switch to the hotbar? I think at this point a better solution would be to make a custom screen handler.

xpple:
I tried, and failed. I created a screen handler that `extends ScreenHandler`, but even like that I wasn't able to prevent the player from dropping items with Q or swapping the selected slot with the number keys. Maybe if you have a look at it you could come up with something but I couldn't.
Stupidly, this seems to do absolutely nothing to prevent it:
```java
@Override
public boolean keyPressed(int keyCode, int scanCode, int modifiers) {
    return keyCode == 27;
}
```

Earthcomputer:
I suppose this is an improvement over what we have right now. We can rewrite the screen handler later.

PR #266, opened 2021-05-20
Add the /cplayerinfo command

### LINKED ISSUE #257: Player Name History Command
I would like a QOL command that will list the name history of someone in chat (privately).
The command may be /cnamehistory or similar. The ability to be able to tab through online players would be appreciated.

xpple:
Closes #257


xpple:
I’ve noticed a stupid mistake in the name of the class. Just letting know I’m aware of the issue.

Earthcomputer (changes requested):
A few things. Firstly, the client already knows other players' UUIDs, so there is no need to request it from the Mojang API if that player is online, you could add a check for that.
Secondly, this API is rate limited, so you should cache the results.
It also may take some time for the server to respond, so you might want to use a CompletableFuture or something and print the result when it arrives, rather than freezing the game until the result arrives.

Finally, in the future we could add more than name history, so maybe this command could use a different format:
`/cplayerinfo [<name>|<uuid>] namehistory`

src/main/java/net/earthcomputer/clientcommands/command/PlayerInfoCommand.java
```diff
@@ -0,0 +1,104 @@
...
+                if (result == null) {
+                    throw new IOException();
+                }
+                return result;
+            } catch (Exception e) {
+                throw new CompletionException(e);
+            }
+        });
+        try {
+            return future.join();
```
Earthcomputer: This isn't actually async because it causes the main thread to wait until the task is completed. The future should be stored somewhere and queried every tick until it's finished

Earthcomputer:
Postponing to use the `HttpRequest` API from Java 16

xpple:
The "postponed" tag can be removed now (as for most other issues/PRs marked with "postponed").

src/main/java/net/earthcomputer/clientcommands/command/PlayerInfoCommand.java
```diff
@@ -0,0 +1,106 @@
...
+                fetchNameHistory(player);
+            }
+        } else {
+            if (cacheByName.containsKey(player)) {
+                sendFeedback(new TranslatableText("commands.cplayerinfo.getNameHistory.success", player, String.join(", ", cacheByName.get(player))));
+            } else {
+                Optional<String> optional = MinecraftClient.getInstance().world.getPlayers().stream()
+                        .filter(abstractPlayer -> abstractPlayer.getName().getString().equals(player))
+                        .map(Entity::getUuidAsString)
+                        .findFirst();
```
Earthcomputer: Might be better to use the PlayerList from the NetworkHandler instead

src/main/java/net/earthcomputer/clientcommands/command/PlayerInfoCommand.java
```diff
@@ -0,0 +1,106 @@
...
+                if (optional.isPresent()) {
+                    fetchNameHistory(optional.get());
+                } else {
+                    HttpClient client = HttpClient.newHttpClient();
```
Earthcomputer: This section could probably be extracted to its own method for clarity

src/main/java/net/earthcomputer/clientcommands/command/PlayerInfoCommand.java
```diff
@@ -0,0 +1,106 @@
...
+                    HttpRequest request = HttpRequest.newBuilder(URI.create("https://api.mojang.com/users/profiles/minecraft/" + player))
+                            .timeout(Duration.ofSeconds(5))
```
Earthcomputer: extract this number 5 into a static final constant? Same below

src/main/java/net/earthcomputer/clientcommands/command/PlayerInfoCommand.java
```diff
@@ -0,0 +1,106 @@
...
+                            .build();
+                    client.sendAsync(request, HttpResponse.BodyHandlers.ofString())
+                            .thenApply(HttpResponse::body)
+                            .thenAccept(response -> {
+                                JsonElement result = parser.parse(response);
+                                if (result instanceof JsonNull) {
+                                    sendError(new TranslatableText("commands.cplayerinfo.ioException"));
+                                } else {
+                                    fetchNameHistory(result.getAsJsonObject().get("id").getAsString());
+                                }
```
Earthcomputer: This will not run on the main thread, leading to thread safety issues when printing to chat. The body of this lambda needs to be wrapped in `MinecraftClient.getInstance().send` to schedule it to be run on the main thread. Same below

src/main/java/net/earthcomputer/clientcommands/command/PlayerInfoCommand.java
```diff
@@ -0,0 +1,106 @@
...
+                            });
+                }
+            }
+        }
+        return 0;
+    }
+
+    private static void fetchNameHistory(String uuid) {
+        HttpClient client = HttpClient.newHttpClient();
```
Earthcomputer: I have no experience with the `HttpClient` API, but would it be better to create one client and reuse it?

Earthcomputer (changes requested):

src/main/java/net/earthcomputer/clientcommands/command/PlayerInfoCommand.java
xpple: Though you never said it was, I wouldn't consider this `5` a magic number; it's quite clear what it means.

Earthcomputer: My issue with it is more that it's duplicated in two places, we might later decide to change it and forget to change it in the other places, that won't happen if it's a constant

src/main/java/net/earthcomputer/clientcommands/command/PlayerInfoCommand.java
xpple: Yes it would be.

src/main/java/net/earthcomputer/clientcommands/command/PlayerInfoCommand.java
xpple: Alright.

src/main/java/net/earthcomputer/clientcommands/command/PlayerInfoCommand.java
xpple: This crashes the client on Singleplayer.

Earthcomputer: What leads you to that conclusion?

Earthcomputer: Okay seems like it doesn't *crash*, but it doesn't find any players either. So I propose you test if it's singleplayer, and on singleplayer get the list directly from the integrated server, and in multiplayer get the list from the network handler.

Earthcomputer (approved):

PR #276, opened 2021-06-07
`/cfindblock` run clook after clicking coordinates.

### LINKED ISSUE #275: Can I have clook when clicking on coordinates in cfindblock?
I'd like an way to have clook instead of, or in addition to, cglow when clicking on a block coordinate in cfindblock, like how some old versions used to do :-(

L0um15:
Closes #275 
Also display [Glow] component next to coordinates.
image

### DIFF
--- a/src/main/resources/assets/clientcommands/lang/en_us.json
+++ b/src/main/resources/assets/clientcommands/lang/en_us.json
@@ -42,7 +42,7 @@
 
   "commands.cfindblock.notFound": "No such block found",
   "commands.cfindblock.success.left": "Closest match is at ",
-  "commands.cfindblock.success.glow": " [Glow]",
+  "commands.cfindblock.success.glow": "[Glow]",
   "commands.cfindblock.success.right": ", %1$.2f blocks away",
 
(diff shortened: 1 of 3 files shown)


src/main/resources/assets/clientcommands/lang/en_us.json
```diff
@@ -42,6 +42,7 @@
 
   "commands.cfindblock.notFound": "No such block found",
   "commands.cfindblock.success.left": "Closest match is at ",
+  "commands.cfindblock.success.glow": " [Glow]",
```
Earthcomputer: Maybe remove the space from here and add it to the code?

L0um15: Sure

L0um15:
Now separator between coordinates and glow component is visible
image

Earthcomputer:
That's fine, thanks!

Earthcomputer (approved):

sudofox:
For some reason it doesn't run the /clook as a client command when I click the coordinates, only when I type it in manually.

PR #296, opened 2021-07-19
Add the /ctp command

### LINKED ISSUE #293: [Feature Request] /ctp only at spectator mode
So, you know that spectator mode can tp to the player. Now, it will be cool if you can tp to specific coordinate with spectator mode. **I don't know if this possible or not**. If not possible, maybe just add the feature without coordinate (only tp to player)

xpple:
Closes #293.

Not only the UUIDs suggested by the client, but also any valid UUID of an entity will be valid as an argument to the command. Because UUIDs alone are quite meaningless, entity selectors are supported as well. For example, use `/ctp @e[type=minecraft:zombie, sort=furthest, limit=1]` to teleport to the furthest zombie known to the client. Sadly, one can't teleport to arbitrary coordinates.

### DIFF
--- a/src/main/java/net/earthcomputer/clientcommands/command/CTeleportCommand.java
+++ b/src/main/java/net/earthcomputer/clientcommands/command/CTeleportCommand.java
@@ -12,14 +12,11 @@ import net.minecraft.text.TranslatableText;
 import java.util.UUID;
 import java.util.stream.StreamSupport;
 
-import static net.earthcomputer.clientcommands.command.ClientCommandManager.addClientSideCommand;
-import static net.earthcomputer.clientcommands.command.arguments.ClientEntityArgumentType.entity;
-import static net.earthcomputer.clientcommands.command.arguments.ClientEntityArgumentType.getEntity;
-import static net.minecraft.command.CommandSource.suggestMatching;
-import static net.minecraft.command.argument.UuidArgumentType.getUuid;
-import static net.minecraft.command.argument.UuidArgumentType.uuid;
-import static net.minecraft.server.command.CommandManager.argument;
-import static net.minecraft.server.command.CommandManager.literal;
+import static net.earthcomputer.clientcommands.command.ClientCommandManager.*;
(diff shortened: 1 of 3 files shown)


src/main/java/net/earthcomputer/clientcommands/command/CTeleportCommand.java
```diff
@@ -0,0 +1,47 @@
...
+    private static final MinecraftClient client = MinecraftClient.getInstance();
+
+    public static void register(CommandDispatcher<ServerCommandSource> dispatcher) {
+        addClientSideCommand("ctp");
+
+        dispatcher.register(literal("ctp")
+                .then(argument("uuid", uuid())
+                        .suggests((context, builder) -> suggestMatching(client.world.getPlayers().stream().map(Entity::getUuidAsString), builder))
+                        .suggests((context, builder) -> suggestMatching(StreamSupport.stream(client.world.getEntities().spliterator(), false).map(Entity::getUuidAsString), builder))
+                        .executes(ctx -> teleport(ctx.getSource(), getUuid(ctx, "uuid"))))
```
Earthcomputer: As of the latest version of clientcommands, the `uuid` arguemnt is unnecessary, as we fixed `EntityArgumentType` to allow uuids

Earthcomputer (changes requested):
Looks good other than this one thing

src/main/java/net/earthcomputer/clientcommands/command/CTeleportCommand.java
xpple: What fix exactly?

Earthcomputer: https://github.com/Earthcomputer/clientcommands/commit/e29ecfc447fa07d62927005d4fd2da931b5a4a81
https://github.com/Earthcomputer/clientcommands/commit/e12288977aaae874e12b42774f6822dbf072255e

Earthcomputer (approved):

PR #302, opened 2021-08-01
Add /calias command 

JoakimThorsen:
Command to create aliases for long/repetitive/complex commands. 

Usage examples:
`/calias add gmc /gamemode creative`
`/gmc`
 
`/calias add glowIF /cglow entities @e[type=minecraft:item_frame,nbt={Item:{id:"minecraft:%s"}}]`
`/glowIF diamond` 

Translation strings not yet implemented, might need some more error handling, planning to append trailing command arguments to the end of the command (or perhaps another implementation to deal with trailing arguments? they currently get ignored), automatically create alias_list.json if missing, auto-add '/' prefix?


src/main/java/net/earthcomputer/clientcommands/command/AliasCommand.java
```diff
@@ -0,0 +1,174 @@
...
+                .then(argument("key", string())
+                    .then(argument("command", greedyString())
+                        .executes(ctx -> addAlias(getString(ctx, "key"), getString(ctx, "command"))))))
+            .then(literal("list")
+                .executes(ctx -> listAliases()))
+            .then(literal("remove")
+                .then(argument("key", string())
+                    .executes(ctx -> removeAlias(getString(ctx, "key"))))));
+
+        HashMap<String, String> aliasMap = getAliases();
```
xpple: I think this code should be put in a `static` block.

src/main/java/net/earthcomputer/clientcommands/command/AliasCommand.java
```diff
@@ -0,0 +1,174 @@
...
+    }
+    private static int executeAliasCommand(String aliasKey, String arguments) {
+
+        // TODO: add support for optional greedy text arguments(?)
+        HashMap<String, String> aliasMap = getAliases();
+        String cmd;
+        if(aliasMap.containsKey(aliasKey)) {
+            cmd = aliasMap.get(aliasKey);
+        } else {
+            sendError(new LiteralText("No such key exists"));
```
xpple: You should probably throw an exception rather than sending an error message.

src/main/java/net/earthcomputer/clientcommands/command/AliasCommand.java
```diff
@@ -0,0 +1,174 @@
...
+        if(arguments!=null) {
+            cmd = String.format(cmd, arguments.split(" "));
+        }
+        MinecraftClient.getInstance().player.sendChatMessage(cmd);
+
+        return 0;
+    }
+
+    private static int addAlias(String key, String command) {
+        ClientPlayNetworkHandler networkHandler = MinecraftClient.getInstance().getNetworkHandler();
```
xpple, line 99: Rather than using `assert`, use an `if`-statement and throw an exception if `null`.

src/main/java/net/earthcomputer/clientcommands/command/AliasCommand.java
```diff
@@ -0,0 +1,174 @@
...
+        }
+        addClientSideCommand(key);
+
+        dispatcher.register(literal(key)
+                .executes(ctx -> executeAliasCommand(key, null))
+                .then(argument("arguments", greedyString())
+                        .executes(ctx -> executeAliasCommand(key, getString(ctx, "arguments")))));
+        aliasMap.put(key, command);
+
+        try {
```
xpple: I think you can use a try-with-resources-statement here.

src/main/java/net/earthcomputer/clientcommands/command/AliasCommand.java
```diff
@@ -0,0 +1,174 @@
...
+            Writer writer = new FileWriter(String.valueOf(aliasPath));
+            Gson gson = new Gson();
+
+            gson.toJson(aliasMap, writer);
+            writer.flush();
+            writer.close();
+
+            sendFeedback(new LiteralText("New alias added successfully!"));
```
xpple: User feedback should always be translatable.

src/main/java/net/earthcomputer/clientcommands/command/AliasCommand.java
```diff
@@ -0,0 +1,174 @@
...
+    private static int addAlias(String key, String command) {
+        ClientPlayNetworkHandler networkHandler = MinecraftClient.getInstance().getNetworkHandler();
+        assert networkHandler != null;
+        var dispatcher = (CommandDispatcher<ServerCommandSource>) (CommandDispatcher<?>) networkHandler.getCommandDispatcher();
+
+        HashMap<String, String> aliasMap = getAliases();
+
+        assert aliasMap != null;
+        if(aliasMap.containsKey(key)) {
+            sendError(new LiteralText("That key already exists!"));
```
xpple: You should probably throw an exception rather than sending an error message.

src/main/java/net/earthcomputer/clientcommands/command/AliasCommand.java
```diff
@@ -0,0 +1,174 @@
...
+            Writer writer = new FileWriter(String.valueOf(aliasPath));
+            Gson gson = new Gson();
+
+            gson.toJson(aliasMap, writer);
+            writer.flush();
+            writer.close();
+
+            sendFeedback(new LiteralText("New alias added successfully!"));
+        } catch (Exception e) {
+            sendError(new LiteralText("Uh oh"));
```
xpple: You should probably throw an exception rather than sending an error message.

src/main/java/net/earthcomputer/clientcommands/command/AliasCommand.java
```diff
@@ -0,0 +1,174 @@
...
+        }
+        return 0;
+    }
+    private static int listAliases() {
+        HashMap<String, String> aliasMap = getAliases();
+        if (aliasMap == null || aliasMap.isEmpty()) {
+            sendError(new LiteralText("No aliases registered"));
```
xpple: You should probably throw an exception rather than sending an error message.

src/main/java/net/earthcomputer/clientcommands/command/AliasCommand.java
```diff
@@ -0,0 +1,174 @@
...
+        assert networkHandler != null;
+        var dispatcher = (CommandDispatcher<ServerCommandSource>) (CommandDispatcher<?>) networkHandler.getCommandDispatcher();
+
+        HashMap<String, String> aliasMap = getAliases();
+
+        if(aliasMap != null && aliasMap.containsKey(key)) {
+            BrigadierRemover.of(dispatcher).get(key).remove();
+            aliasMap.remove(key);
+        } else {
+            sendError(new LiteralText("No such key exists"));
```
xpple: You should probably throw an exception rather than sending an error message.

src/main/java/net/earthcomputer/clientcommands/command/AliasCommand.java
```diff
@@ -0,0 +1,174 @@
...
+        try {
+            Writer writer = new FileWriter(String.valueOf(aliasPath));
+            Gson gson = new Gson();
+
+            gson.toJson(aliasMap, writer);
+            writer.flush();
+            writer.close();
+            sendFeedback(new LiteralText("Alias removed successfully!"));
+        } catch (Exception e) {
+            sendError(new LiteralText("Uh oh"));
```
xpple: You should probably throw an exception rather than sending an error message.

src/main/java/net/earthcomputer/clientcommands/command/AliasCommand.java
```diff
@@ -0,0 +1,174 @@
...
+    private static HashMap<String, String> getAliases() {
+
+        LOGGER.info(aliasPath.toString());
+
+        Gson gson = new Gson();
+        FileReader fileReader = null;
+        try {
+            fileReader = new FileReader(String.valueOf(aliasPath));
+        } catch (Exception e) {
+            sendError(new LiteralText("Failed to read alias file."));
```
xpple: You should probably throw an exception rather than sending an error message.

src/main/java/net/earthcomputer/clientcommands/command/AliasCommand.java
```diff
@@ -0,0 +1,174 @@
...
+            .then(literal("list")
+                .executes(ctx -> listAliases()))
+            .then(literal("remove")
+                .then(argument("key", string())
+                    .executes(ctx -> removeAlias(getString(ctx, "key"))))));
+
+        HashMap<String, String> aliasMap = getAliases();
+
+        if (aliasMap == null || aliasMap.isEmpty()) {
+            sendError(new LiteralText("No aliases registered"));
```
xpple: You should probably throw an exception rather than sending an error message.

src/main/java/net/earthcomputer/clientcommands/command/AliasCommand.java
```diff
@@ -0,0 +1,174 @@
...
+            sendError(new LiteralText("Uh oh"));
+        }
+        return 0;
+    }
+    private static int listAliases() {
+        HashMap<String, String> aliasMap = getAliases();
+        if (aliasMap == null || aliasMap.isEmpty()) {
+            sendError(new LiteralText("No aliases registered"));
+        } else {
+            sendFeedback(new LiteralText("Aliases registered: " + aliasMap.size()));
```
xpple: User feedback should always be translatable.

src/main/java/net/earthcomputer/clientcommands/command/AliasCommand.java
```diff
@@ -0,0 +1,174 @@
...
+        }
+
+        try {
+            Writer writer = new FileWriter(String.valueOf(aliasPath));
+            Gson gson = new Gson();
+
+            gson.toJson(aliasMap, writer);
+            writer.flush();
+            writer.close();
+            sendFeedback(new LiteralText("Alias removed successfully!"));
```
xpple: User feedback should always be translatable.

src/main/java/net/earthcomputer/clientcommands/command/AliasCommand.java
```diff
@@ -0,0 +1,174 @@
...
+        if(aliasMap != null && aliasMap.containsKey(key)) {
+            BrigadierRemover.of(dispatcher).get(key).remove();
+            aliasMap.remove(key);
+        } else {
+            sendError(new LiteralText("No such key exists"));
+            return 0;
+        }
+
+        try {
+            Writer writer = new FileWriter(String.valueOf(aliasPath));
```
xpple: I think you can use a try-with-resources-statement here.

src/main/java/net/earthcomputer/clientcommands/command/AliasCommand.java
```diff
@@ -0,0 +1,174 @@
...
+        return 0;
+    }
+
+    private static HashMap<String, String> getAliases() {
+
+        LOGGER.info(aliasPath.toString());
+
+        Gson gson = new Gson();
+        FileReader fileReader = null;
+        try {
```
xpple: I think you can use a try-with-resources-statement here.

src/main/java/net/earthcomputer/clientcommands/command/AliasCommand.java
```diff
@@ -0,0 +1,174 @@
...
+    }
+
+    private static int addAlias(String key, String command) {
+        ClientPlayNetworkHandler networkHandler = MinecraftClient.getInstance().getNetworkHandler();
+        assert networkHandler != null;
+        var dispatcher = (CommandDispatcher<ServerCommandSource>) (CommandDispatcher<?>) networkHandler.getCommandDispatcher();
+
+        HashMap<String, String> aliasMap = getAliases();
+
+        assert aliasMap != null;
```
xpple: Rather than using `assert`, use an `if`-statement and throw an exception if `null`.

src/main/java/net/earthcomputer/clientcommands/command/AliasCommand.java
```diff
@@ -0,0 +1,174 @@
...
+            for(String key: aliasMap.keySet()) {
+                sendFeedback(Formatting.BOLD + key + Formatting.RESET+ ": "+ aliasMap.get(key).replace("%","%%"));
+            }
+
+        }
+        return 0;
+    }
+    private static int removeAlias(String key) {
+        ClientPlayNetworkHandler networkHandler = MinecraftClient.getInstance().getNetworkHandler();
+        assert networkHandler != null;
```
xpple, line 138: Rather than using `assert`, use an `if`-statement and throw an exception if `null`.

xpple (changes requested):
Thanks for the pull request!

There might be some code-style things I haven’t addressed, I’ll have to wait for Earth’s opinion on that.

src/main/java/net/earthcomputer/clientcommands/command/AliasCommand.java
Earthcomputer: `assert` is used for things that you don't expect *could* ever happen, like here. It's also to make intellij's nullability checker happy

src/main/java/net/earthcomputer/clientcommands/command/AliasCommand.java
xpple: Actually, this line can be removed entirely after applying the suggested changes in the `getAliases` method.

src/main/java/net/earthcomputer/clientcommands/command/AliasCommand.java
mariusdkm: Since this is user feedback, that an alias already exists, I don't think an exception should be thrown.
And I also think, that if an alias already exists the old alias should be overwritten.

src/main/java/net/earthcomputer/clientcommands/command/AliasCommand.java
mariusdkm: I also think here, since the user specifically requested to list the aliases and no aliases exist, some kind of feedback is better than throwing an exception.
(But the feedback should be translatable)

src/main/java/net/earthcomputer/clientcommands/command/AliasCommand.java
mariusdkm: Again here, I think this is user feedback, so no exception should be thrown.

src/main/java/net/earthcomputer/clientcommands/command/AliasCommand.java
JoakimThorsen: Agreed. "No aliases registered" is more informative to the user than "An unexpected error occured" which is in this case the alternative

src/main/java/net/earthcomputer/clientcommands/command/AliasCommand.java
JoakimThorsen: I like the suggestion of overwriting the existing alias, perhaps with a clickable [undo?] button in chat if the alias was overwritten by accident

src/main/java/net/earthcomputer/clientcommands/command/AliasCommand.java
JoakimThorsen: This is the only one I might agree with, as this one could at least be handled elsewhere

src/main/java/net/earthcomputer/clientcommands/command/AliasCommand.java
xpple: Following the style of `ckit` and `citemgroup`, an `ALREADY_EXISTS_EXCEPTION` should be thrown.

src/main/java/net/earthcomputer/clientcommands/command/AliasCommand.java
xpple: Notice that a thrown exception also sends feedback to the client. The only difference is throwing the exception as opposed to sending an error message directly. In this case, I think the `if`-statement should be split in two. First check for `null` and throw an exception if so, then check `isEmpty` and send proper feedback if true.

src/main/java/net/earthcomputer/clientcommands/command/AliasCommand.java
xpple: If it’s actually feedback, I think `sendFeedback` should be used. I always try to avoid `sendError`, and throw an exception instead.
This style is also documented here: [CONTRIBUTING.md](https://github.com/Earthcomputer/clientcommands/blob/fabric/CONTRIBUTING.md).
The only time I used `sendError` was when I couldn’t possibly throw an exception (`cplayerinfo`).

JoakimThorsen:
New commit, most issues so far should be sorted. 
- Still need to find out what to do with trailing arguments
- Haven't decided on whether or not to do `/`-prefixing
- String variables(?)

src/main/java/net/earthcomputer/clientcommands/command/AliasCommand.java
```diff
@@ -0,0 +1,164 @@
...
+            }
+        }
+        return 0;
+    }
+    private static int removeAlias(String key) throws CommandSyntaxException {
+        ClientPlayNetworkHandler networkHandler = MinecraftClient.getInstance().getNetworkHandler();
+        assert networkHandler != null;
+        var dispatcher = (CommandDispatcher<ServerCommandSource>) (CommandDispatcher<?>) networkHandler.getCommandDispatcher();
+
+        if(aliasMap != null && aliasMap.containsKey(key)) {
```
xpple: In think this `if`-statement should be split in two.

src/main/java/net/earthcomputer/clientcommands/command/AliasCommand.java
```diff
@@ -0,0 +1,164 @@
...
+                                .executes(ctx -> executeAliasCommand(key, getString(ctx, "arguments")))));
+            }
+        }
+
+    }
+    private static int executeAliasCommand(String aliasKey, String arguments) throws CommandSyntaxException {
+
+        // TODO: add support for optional greedy text arguments(?)
+        String cmd;
+        if(aliasMap != null && aliasMap.containsKey(aliasKey)) {
```
xpple: I think this `if`-statement should be split in two.

src/main/java/net/earthcomputer/clientcommands/command/AliasCommand.java
```diff
@@ -0,0 +1,164 @@
...
+
+        if (aliasMap != null && !aliasMap.isEmpty()) {
+            for(String key: aliasMap.keySet()) {
+                addClientSideCommand(key);
+                dispatcher.register(literal(key)
+                        .executes(ctx -> executeAliasCommand(key, null))
+                        .then(argument("arguments", greedyString())
+                                .executes(ctx -> executeAliasCommand(key, getString(ctx, "arguments")))));
+            }
+        }
```
xpple: This code should also be inside the `static` block.

src/main/java/net/earthcomputer/clientcommands/command/AliasCommand.java
```diff
@@ -0,0 +1,164 @@
...
+import java.nio.file.Path;
+import java.util.HashMap;
+
+import static com.mojang.brigadier.arguments.StringArgumentType.*;
+import static net.earthcomputer.clientcommands.command.ClientCommandManager.*;
+import static net.minecraft.server.command.CommandManager.*;
+
+public class AliasCommand {
+
+    private static final Logger LOGGER = LogManager.getLogger();
```
xpple: The logger should have the name `clientcommands`.

xpple (changes requested):
Nearly there!

src/main/java/net/earthcomputer/clientcommands/command/AliasCommand.java
JoakimThorsen: I would disagree, but it seems the first half can be removed entirely. I don't see a point in throwing a unique exception here.

src/main/java/net/earthcomputer/clientcommands/command/AliasCommand.java
```diff
@@ -38,10 +38,24 @@
...
+    private static CommandDispatcher<ServerCommandSource> dispatcher;
 
     static {
         try {
+            ClientPlayNetworkHandler networkHandler = MinecraftClient.getInstance().getNetworkHandler();
+            assert networkHandler != null;
+            dispatcher = (CommandDispatcher<ServerCommandSource>) (CommandDispatcher<?>) networkHandler.getCommandDispatcher();
+
             getAliases();
+            if (!aliasMap.isEmpty()) {
```
xpple: Are you sure this `if`-statement is needed?

xpple (changes requested):
Thanks! Just one minor request. I’ll have to check with Earth about some of the code-style choices.

src/main/java/net/earthcomputer/clientcommands/command/AliasCommand.java
JoakimThorsen: ah, probably not, good call

src/main/java/net/earthcomputer/clientcommands/command/AliasCommand.java
```diff
@@ -0,0 +1,178 @@
...
+    private static final DynamicCommandExceptionType NOT_FOUND_EXCEPTION = new DynamicCommandExceptionType(arg -> new TranslatableText("commands.calias.notFound", arg));
+
+    private static HashMap<String, String> aliasMap = new HashMap<>();
+
+    static {
+        try {
+            getAliases();
+        } catch (Exception e) {
+            e.printStackTrace();
+            LOGGER.info("No alias file provided. A new one will be created upon registering an alias.");
```
xpple: I think you should use `Logger#error` here and append the exception as a parameter to the method.

src/main/java/net/earthcomputer/clientcommands/command/AliasCommand.java
```diff
@@ -0,0 +1,178 @@
...
+
+        String cmd;
+        if (aliasMap.containsKey(aliasKey)) {
+            cmd = aliasMap.get(aliasKey);
+        } else {
+            throw NOT_FOUND_EXCEPTION.create(aliasKey);
+        }
+        int inlineArgumentCount = (int) Pattern.compile("%[^%]").matcher(cmd).results().count();
+        if (inlineArgumentCount > 0) {
+            String[] argumentArray = arguments.split(" ", inlineArgumentCount+1);
```
xpple: Could you surround the `+` with spaces?

xpple (changes requested):
Looks good! I'll leave the actual code-review to Earth.

src/main/java/net/earthcomputer/clientcommands/command/AliasCommand.java
```diff
@@ -0,0 +1,186 @@
...
+
+    private static final SimpleCommandExceptionType FILE_READ_EXCEPTION = new SimpleCommandExceptionType(new TranslatableText("commands.calias.file.readError"));
+    private static final SimpleCommandExceptionType FILE_WRITE_EXCEPTION = new SimpleCommandExceptionType(new TranslatableText("commands.calias.file.writeError"));
+    private static final SimpleCommandExceptionType TOO_FEW_ARGUMENTS_EXCEPTION = new SimpleCommandExceptionType(new TranslatableText("commands.calias.tooFewArguments"));
+
+    private static final DynamicCommandExceptionType ALIAS_EXISTS_EXCEPTION = new DynamicCommandExceptionType(arg -> new TranslatableText("commands.calias.addAlias.aliasAlreadyExists", arg));
+    private static final DynamicCommandExceptionType COMMAND_EXISTS_EXCEPTION = new DynamicCommandExceptionType(arg -> new TranslatableText("commands.calias.addAlias.commandAlreadyExists", arg));
+    private static final DynamicCommandExceptionType NOT_FOUND_EXCEPTION = new DynamicCommandExceptionType(arg -> new TranslatableText("commands.calias.notFound", arg));
+
+    private static HashMap<String, String> aliasMap = new HashMap<>();
```
Earthcomputer: This field can be final

src/main/java/net/earthcomputer/clientcommands/command/AliasCommand.java
```diff
@@ -0,0 +1,186 @@
...
+        }
+
+        saveAliases();
+        sendFeedback(new TranslatableText("commands.calias.removeAlias.success", key));
+        return 0;
+    }
+
+    private static void getAliases() throws CommandSyntaxException {
+
+        LOGGER.info("Alias config path:" + aliasPath.toString());
```
Earthcomputer: I don't think we need to log this, it's just noise in the log file

src/main/java/net/earthcomputer/clientcommands/command/AliasCommand.java
```diff
@@ -0,0 +1,186 @@
...
+        } else {
+            throw NOT_FOUND_EXCEPTION.create(key);
+        }
+
+        saveAliases();
+        sendFeedback(new TranslatableText("commands.calias.removeAlias.success", key));
+        return 0;
+    }
+
+    private static void getAliases() throws CommandSyntaxException {
```
Earthcomputer: methods named `getXXX` sort of imply that it's a getter in Java, this is not a getter. Could be renamed to `loadAliases`.

src/main/java/net/earthcomputer/clientcommands/command/AliasCommand.java
```diff
@@ -0,0 +1,186 @@
...
+
+        LOGGER.info("Alias config path:" + aliasPath.toString());
+
+        Gson gson = new Gson();
+        try (FileReader fileReader = new FileReader(String.valueOf(aliasPath))) {
+            aliasMap = gson.fromJson(new JsonReader(fileReader), HashMap.class);
+        } catch (Exception e) {
+            throw FILE_READ_EXCEPTION.create();
```
Earthcomputer: Throwing a `CommandSyntaxException` is usually for if this is run in a command, as Minecraft catches and reports those exceptions to chat. Since this is called in the `static {}` block you could directly log the actual exception here and not throw it.

src/main/java/net/earthcomputer/clientcommands/command/AliasCommand.java
```diff
@@ -0,0 +1,186 @@
...
+        assert networkHandler != null;
+        var dispatcher = (CommandDispatcher<ServerCommandSource>) (CommandDispatcher<?>) networkHandler.getCommandDispatcher();
+
+        if (aliasMap.containsKey(key)) {
+            throw ALIAS_EXISTS_EXCEPTION.create(key);
+        }
+        if (dispatcher.getRoot().getChildren().stream().map(CommandNode::getName).anyMatch(literal -> literal.equals(key))) {
+            throw COMMAND_EXISTS_EXCEPTION.create(key);
+        }
+        if (command.charAt(0) != '/') {
```
Earthcomputer:
```suggestion
        if (!command.startsWith("/")) {
```
this also handles the case where `command` is empty

src/main/java/net/earthcomputer/clientcommands/command/AliasCommand.java
```diff
@@ -0,0 +1,186 @@
...
+        }
+    }
+    private static int executeAliasCommand(String aliasKey, String arguments) throws CommandSyntaxException {
+
+        String cmd;
+        if (aliasMap.containsKey(aliasKey)) {
+            cmd = aliasMap.get(aliasKey);
+        } else {
+            throw NOT_FOUND_EXCEPTION.create(aliasKey);
+        }
```
Earthcomputer:
```suggestion
        String cmd = aliasMap.get(aliasKey);
        if (cmd == null) {
            throw NOT_FOUND_EXCEPTION.create(aliasKey);
        }
```

src/main/java/net/earthcomputer/clientcommands/command/AliasCommand.java
```diff
@@ -0,0 +1,186 @@
...
+        int inlineArgumentCount = (int) Pattern.compile("%[^%]").matcher(cmd).results().count();
```
Earthcomputer:
```suggestion
        int inlineArgumentCount = (int) Pattern.compile("(?<!%)%(?:%%)*(?!%)").matcher(cmd).results().count();
```
Also you should probably move this pattern into a field so it doesn't have to get recompiled each time

src/main/java/net/earthcomputer/clientcommands/command/AliasCommand.java
```diff
@@ -0,0 +1,186 @@
...
+            String[] argumentArray = arguments.split(" ", inlineArgumentCount + 1);
+
+            String trailingArguments = "";
+            if (argumentArray.length > inlineArgumentCount) {
+                trailingArguments = " " + argumentArray[inlineArgumentCount];
+            }
+            try {
+                cmd = String.format(cmd, (Object[]) argumentArray) + trailingArguments;
+            } catch (Exception e) {
+                LOGGER.error(e);
```
Earthcomputer: Don't log this exception

src/main/java/net/earthcomputer/clientcommands/command/AliasCommand.java
```diff
@@ -0,0 +1,186 @@
...
+                throw TOO_FEW_ARGUMENTS_EXCEPTION.create();
```
Earthcomputer: This could be too few arguments or an illegal format string

src/main/java/net/earthcomputer/clientcommands/command/AliasCommand.java
```diff
@@ -0,0 +1,186 @@
...
+
+import static com.mojang.brigadier.arguments.StringArgumentType.*;
+import static net.earthcomputer.clientcommands.command.ClientCommandManager.*;
+import static net.minecraft.server.command.CommandManager.*;
+
+public class AliasCommand {
+
+    private static final Logger LOGGER = LogManager.getLogger("clientcommands");
+
+    private static final Path aliasPath = FabricLoader.getInstance().getConfigDir().resolve("clientcommands").resolve("alias_list.json");
```
Earthcomputer: This constant should be capitalized

src/main/java/net/earthcomputer/clientcommands/command/AliasCommand.java
```diff
@@ -0,0 +1,186 @@
...
+                        .executes(ctx -> executeAliasCommand(key, null))
+                        .then(argument("arguments", greedyString())
+                                .executes(ctx -> executeAliasCommand(key, getString(ctx, "arguments")))));
+            } else {
+                LOGGER.error("Attempted to register alias /{}, but that command already exists", key);
+            }
+        }
+    }
+    private static int executeAliasCommand(String aliasKey, String arguments) throws CommandSyntaxException {
+
```
Earthcomputer: Not usually this picky about formatting but this really bugs me. Please no blank lines at the start of methods

src/main/java/net/earthcomputer/clientcommands/command/AliasCommand.java
```diff
@@ -0,0 +1,186 @@
...
+        return 0;
+    }
+
+    private static void getAliases() throws CommandSyntaxException {
+
+        LOGGER.info("Alias config path:" + aliasPath.toString());
+
+        Gson gson = new Gson();
+        try (FileReader fileReader = new FileReader(String.valueOf(aliasPath))) {
+            aliasMap = gson.fromJson(new JsonReader(fileReader), HashMap.class);
```
Earthcomputer:
```suggestion
            aliasMap = gson.fromJson(new JsonReader(fileReader), new TypeToken<HashMap<String, String>>(){}.getType());
```

Earthcomputer (changes requested):

src/main/java/net/earthcomputer/clientcommands/command/AliasCommand.java
JoakimThorsen: It was previously a getter. Will update

src/main/java/net/earthcomputer/clientcommands/command/AliasCommand.java
JoakimThorsen: Can it...?

JoakimThorsen: Made this a final and thus reverted to the getter name

src/main/java/net/earthcomputer/clientcommands/ClientCommands.java
```diff
@@ -56,6 +56,8 @@ public static void registerCommands(CommandDispatcher<ServerCommandSource> dispa
         CParticleCommand.register(dispatcher);
         PermissionLevelCommand.register(dispatcher);
         CTimeCommand.register(dispatcher);
+        AliasCommand.register(dispatcher);
+
```
Earthcomputer, line 60: Extra line crept in during the merge

Earthcomputer (approved):
Tyvm!

PR #308, opened 2021-08-07
Add the ctime command

haykam821:


Earthcomputer (approved):
This could be extended later to how the 1.12 clientcommands ctime command was

Issue #313, opened 2021-08-14
Rendering bug around Immersive Portal in 1.17.1


Kouchya:
I was testing this mod in a 1.17.1 single-player world along with the Immersive Portal mod (both newest version for MC 1.17.1). I found that as long as an immersive portal is in sight, some weird xray effect appears around the portal as the following screenshot shows:
![2021-08-14_21 57 32](https://user-images.githubusercontent.com/22111513/129448709-b280878b-97f0-464e-9a77-daec612fa65f.png)

After removing clientcommands it just displays normally, so I guess something might be wrong with the mod's rendering module. Any idea how this happens?

Hexcede:
I think this https://github.com/Earthcomputer/clientcommands/issues/269#issuecomment-853542903 is related

ImUrX:
happens to me too, i was using indium, sodium, cc and immersive, and when i remove cc it stops happening :p

ImUrX:
@Earthcomputer this didn't get closed.

Btw I tested it and it works now

PR #324, opened 2021-10-14
fix and simplify line rendering

### LINKED ISSUE #269: Render issue with shaders
Hi, I have noticed that this mod breaks optifine shaders. I have tried it with complementary shaders and the underwater colour is broken see this [video](https://user-images.githubusercontent.com/46494907/120507339-02558780-c3c7-11eb-8d91-613b5e2c76f7.mp4)

In sildurs vibrant it is even more broken as you can see in this [video](
https://user-images.githubusercontent.com/46494907/120507794-69733c00-c3c7-11eb-862b-089fe5c94eda.mp4)
These are the only mods I had on at that time:
image
Any ideas why this mod would break shaders?
Have a nice day!

wagyourtail:
fixes #269 and #313

Best I can tell, the default Line layer enables depth when put as the active Immediate buffer...
so I set up and teardown/draw the buffers directly using `Tessellator`.


Earthcomputer:
Thanks!

PR #331, opened 2021-11-02
Add the cblockstate command

haykam821:
This pull request adds a command that lists the properties and their values of a specific block state.

`/cblockstate minecraft:oak_slab`

image

`/cblockstate minecraft:oak_slab[type=top,waterlogged=true]`

image


Earthcomputer:
This PR needs updating after the fcapi merge

src/main/java/net/earthcomputer/clientcommands/command/BlockStateCommand.java
```diff
@@ -0,0 +1,57 @@
...
+    private static Text getPropertyLine(Map.Entry<Property<?>, Comparable<?>> entry) {
+        Property<?> property = entry.getKey();
+        Comparable<?> value = entry.getValue();
+
+        MutableText valueText = Text.literal(Util.getValueAsString(property, value));
+        if (Boolean.TRUE.equals(value)) {
+            valueText.formatted(Formatting.GREEN);
+        } else if (Boolean.FALSE.equals(value)) {
+            valueText.formatted(Formatting.RED);
+        }
```
Earthcomputer: I'm pretty sure `Boolean` checks can be replaced with identity comparison

Gaming32: Yes, `Boolean.TRUE` and `Boolean.FALSE` are singletons (or would that be a doubleton 🤔)

haykam821: This check is adapted from the one in `DebugHud`, which also compares the property value to the boxed boolean. Besides, this check needs to handle the case where the property value is not a boolean at all.

haykam821: Fixed in e65aa6dfe4c9a3bfa7defa03f15ea328bbe6772a

src/main/java/net/earthcomputer/clientcommands/command/BlockStateCommand.java
```diff
@@ -0,0 +1,57 @@
...
+        Identifier id = Registry.BLOCK.getId(state.getBlock());
+        String idStr = id == null ? "Unregistered" : id.toString();
+
+        source.sendFeedback(Text.translatable("commands.cblockstate.header", name, idStr, state.getProperties().size()));
+
+        for (Map.Entry<Property<?>, Comparable<?>> entry : state.getEntries().entrySet()) {
+            source.sendFeedback(getPropertyLine(entry));
+        }
+
+        return 0;
```
xpple, line 40:
```suggestion
        return Command.SINGLE_SUCCESS;
```

src/main/java/net/earthcomputer/clientcommands/command/BlockStateCommand.java
```diff
@@ -0,0 +1,57 @@
...
+        dispatcher.register(literal("cblockstate")
+            .then(argument("block", blockState(registryAccess))
+                .executes(ctx -> showBlockState(ctx.getSource(), getCBlockState(ctx, "block").getBlockState()))));
+    }
+
+    private static int showBlockState(FabricClientCommandSource source, BlockState state) {
+        Text name = state.getBlock().getName();
+
+        Identifier id = Registry.BLOCK.getId(state.getBlock());
+        String idStr = id == null ? "Unregistered" : id.toString();
```
xpple, line 32: `id` is never `null`.

src/main/java/net/earthcomputer/clientcommands/command/BlockStateCommand.java
```diff
@@ -0,0 +1,57 @@
+package net.earthcomputer.clientcommands.command;
```
xpple, line 1: The import order is a bit weird.

xpple (changes requested):

PR #350, opened 2022-01-07
Update CONTRIBUTING

xpple:


haykam821:
Could this instead say to target the same Java version as Minecraft? Even though Minecraft 1.18 supports the latest version of Java now, the version supported by Minecraft could be an outdated Java version later.

Earthcomputer:
Yep, or to be more specific "don't use Java features only available in Java versions newer than the one Minecraft uses, if applicable."

PR #351, opened 2022-01-07
Add a "usage tree" command.

wagyourtail:
adds a command which generates trees representing the usage of other commands.
result of `/cusagetree cglow`:
image
image

the green represents that a command can terminate (execute) on that arg

### DIFF
--- a/src/main/java/net/earthcomputer/clientcommands/command/UsageTreeCommand.java
+++ b/src/main/java/net/earthcomputer/clientcommands/command/UsageTreeCommand.java
@@ -4,6 +4,7 @@ import com.google.common.collect.Iterables;
 import com.mojang.brigadier.CommandDispatcher;
 import com.mojang.brigadier.exceptions.SimpleCommandExceptionType;
 import com.mojang.brigadier.tree.CommandNode;
+import net.minecraft.command.CommandSource;
 import net.minecraft.server.command.ServerCommandSource;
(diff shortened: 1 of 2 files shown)


src/main/java/net/earthcomputer/clientcommands/command/UsageTreeCommand.java
```diff
@@ -0,0 +1,92 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+import static com.mojang.brigadier.arguments.StringArgumentType.*;
+import static net.earthcomputer.clientcommands.command.ClientCommandManager.*;
+import static net.minecraft.server.command.CommandManager.*;
+
+public class UsageTreeCommand {
+    private static final SimpleCommandExceptionType FAILED_EXCEPTION = new SimpleCommandExceptionType(new TranslatableText("commands.help.failed"));
```
xpple, line 24: It might be better to use a separate translation string for this command. Arguably the exception name is a bit vague too.

xpple (changes requested):

src/main/java/net/earthcomputer/clientcommands/command/UsageTreeCommand.java
wagyourtail: well, it has the same reason for error as help so...

Earthcomputer (approved):
Looks awesome

PR #353, opened 2022-01-17
add ctellraw

wagyourtail:
image


Earthcomputer:
Ah, it's like /cnote but allows for chat colours. Didn't someone suggest this before? Anyway, thanks for the PR :)

barraIhsan:
> Didn't someone suggest this before?

yes, #208 similar but different

PR #355, opened 2022-01-20
prevent null player from crashing

### LINKED ISSUE #344: More flexible /calias command arguments
#### Suggestion
Add arguments inside a command for /calias, not just at the end
#### Description and Examples
for a command like `/cfind @e[type=minecraft:item,nbt={Item:{id:"minecraft:arrow"}}]` to find an specific item on the ground, we can do something like `/calias item(a) cfind @e[type=minecraft:item,nbt={Item:{id:"minecraft:%a"}}]`, and then to find arrows, we could use `/item("arrow")`
I know this is wrong, but maybe this could be somehow implemented!

Or is it already possible?

### LINKED ISSUE #354: Crash/incompatibility with mine little pony
https://github.com/MineLittlePony/Kirin/issues/4

wagyourtail:
(when other mods run the entity renderer without a world)
fixes #354 (yes I actually checked...)

### DIFF
--- a/src/main/java/net/earthcomputer/clientcommands/features/RenderSettings.java
+++ b/src/main/java/net/earthcomputer/clientcommands/features/RenderSettings.java
@@ -3,6 +3,7 @@ package net.earthcomputer.clientcommands.features;
 import net.earthcomputer.clientcommands.command.ClientEntitySelector;
 import net.earthcomputer.clientcommands.command.FakeCommandSource;
 import net.minecraft.client.MinecraftClient;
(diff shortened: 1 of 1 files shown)


Earthcomputer:
Maybe add a comment explaining under what circumstances the player could be null?

wagyourtail:
:+1:

sudofox:
I, somehow, ran into the exact same issue with Mine Little Pony, ten minutes after discovering its existence.

Earthcomputer (approved):

PR #373, opened 2022-03-18
Add `exactly` parameter

xpple:
Adds the ability to terminate the enchantment argument with the `exactly` parameter. Previously users would have to add `without <enchantment> *` for each enchantment they did not want. Using `exactly` to terminate the argument will logically ignore all occurrences of `without <enchantment> <level>`.
Edit: I should also note I have cleaned up some of the old code to comply with the styling guidelines.


src/main/resources/assets/clientcommands/lang/en_us.json
```diff
@@ -43,7 +43,7 @@
   "commands.ccheatcrackrng.success": "Player RNG cracked: %d -> %d",
   "commands.ccheatcrackrng.java14": "This command no longer works on Java 14+!",
 
-  "commands.cenchant.expectedWithWithout": "Expected \"with\"/\"without\"",
+  "commands.cenchant.expectedWithWithoutExactly": "Expected \"with\"/\"without\"/\"exactly\"",
```
Earthcomputer, line 44: Removing translations requires them to be removed in all the different language files. When I change a translation like this, I usually just delete the translated ones and let the translators re-add it later.

Earthcomputer (changes requested):
Overall very good quality PR. Just one thing

Earthcomputer (approved):

PR #394, opened 2022-04-15
Add `/ctime set` command

Packsolite:
This PR adds a `/ctime set` command allowing to change the clients daytime as discussed in the earthcomputer discord earlier. This is client side only and works in single and multiplayer.

**Command usage**
`/ctime set <day/night/noon/midnight>` to lock the time clientside
`/ctime reset` to unlock and reset to the servers time

### DIFF
--- a/src/main/java/net/earthcomputer/clientcommands/command/CTimeCommand.java
+++ b/src/main/java/net/earthcomputer/clientcommands/command/CTimeCommand.java
@@ -10,8 +10,10 @@ import net.minecraft.client.MinecraftClient;
 import net.minecraft.command.argument.TimeArgumentType;
 import net.minecraft.text.TranslatableText;
 
+import static com.mojang.brigadier.arguments.IntegerArgumentType.getInteger;
 import static net.earthcomputer.clientcommands.command.ClientCommandHelper.*;
 import static net.fabricmc.fabric.api.client.command.v1.ClientCommandManager.*;
+import static net.minecraft.command.argument.TimeArgumentType.time;
 
 public class CTimeCommand {
 
@@ -35,8 +37,8 @@ public class CTimeCommand {
                     .executes(ctx -> executeSetTime(13000)))
                 .then(literal("midnight")
                     .executes(ctx -> executeSetTime(18000)))
-                .then(argument("time", TimeArgumentType.time())
-                    .executes(ctx -> executeSetTime(IntegerArgumentType.getInteger(ctx, "time")))))
+                .then(argument("time", time())
+                    .executes(ctx -> executeSetTime(getInteger(ctx, "time")))))
              .then(literal("reset")
                      .executes(ctx -> executeResetTime())) 
         );
--- a/src/main/java/net/earthcomputer/clientcommands/command/CTimeCommand.java
+++ b/src/main/java/net/earthcomputer/clientcommands/command/CTimeCommand.java
@@ -1,9 +1,13 @@
 package net.earthcomputer.clientcommands.command;
 
 import com.mojang.brigadier.CommandDispatcher;
+import com.mojang.brigadier.arguments.IntegerArgumentType;
+
+import net.earthcomputer.clientcommands.features.ClientTimeModifier;
 import net.fabricmc.fabric.api.client.command.v1.FabricClientCommandSource;
 import net.minecraft.SharedConstants;
 import net.minecraft.client.MinecraftClient;
+import net.minecraft.command.argument.TimeArgumentType;
 import net.minecraft.text.TranslatableText;
 
 import static net.earthcomputer.clientcommands.command.ClientCommandHelper.*;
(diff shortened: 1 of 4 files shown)


Earthcomputer (commented):
Looks good, one more thing, would it be possible to tell the client that doDaylightCycle is false if the type is LOCK? That way, it wouldn't try to interpolate the sun for rendering every tick.
(Don't actually implement it via the game rule of course, but something roughly to that effect)

src/main/java/net/earthcomputer/clientcommands/command/CTimeCommand.java
```diff
@@ -1,13 +1,19 @@
...
 import net.fabricmc.fabric.api.client.command.v1.FabricClientCommandSource;
 import net.minecraft.SharedConstants;
 import net.minecraft.client.MinecraftClient;
+import net.minecraft.command.argument.TimeArgumentType;
 import net.minecraft.text.TranslatableText;
 
+import static com.mojang.brigadier.arguments.IntegerArgumentType.getInteger;
 import static net.earthcomputer.clientcommands.command.ClientCommandHelper.*;
 import static net.fabricmc.fabric.api.client.command.v1.ClientCommandManager.*;
+import static net.minecraft.command.argument.TimeArgumentType.time;
```
haykam821: Prefer using [`CTimeArgument`](https://github.com/xpple/clientarguments/blob/1ade89e78e90aecf4c4a443eaedd0070ab41ba1c/src/main/java/dev/xpple/clientarguments/arguments/CTimeArgumentType.java) to be consistent with other non-Brigadier argument types.

src/main/java/net/earthcomputer/clientcommands/command/CTimeCommand.java
```diff
@@ -1,13 +1,19 @@
 package net.earthcomputer.clientcommands.command;
 
 import com.mojang.brigadier.CommandDispatcher;
+import com.mojang.brigadier.arguments.IntegerArgumentType;
```
haykam821: This import is unused since the `IntegerArgumentType.getInteger` method is imported statically instead.

src/main/java/net/earthcomputer/clientcommands/command/CTimeCommand.java
```diff
@@ -1,13 +1,19 @@
...
+
+import net.earthcomputer.clientcommands.features.ClientTimeModifier;
 import net.fabricmc.fabric.api.client.command.v1.FabricClientCommandSource;
 import net.minecraft.SharedConstants;
 import net.minecraft.client.MinecraftClient;
+import net.minecraft.command.argument.TimeArgumentType;
```
haykam821: This import is unused since the `TimeArgumentType.time` method is imported statically instead.

src/main/java/net/earthcomputer/clientcommands/features/ClientTimeModifier.java
```diff
@@ -0,0 +1,34 @@
...
+    public static void lock(long time) {
+        type = Type.LOCK;
+        value = time;
+    }
+
+    public static long getModifiedTime(long timeOfDay) {
+        return switch (type) {
+            case NONE -> timeOfDay;
+            case LOCK -> value;
+            case OFFSET -> timeOfDay < 0 ? timeOfDay - value : timeOfDay + value;
```
haykam821: Would it make more sense to provide these implementations on the enum members themselves?

haykam821 (changes requested):

src/main/java/net/earthcomputer/clientcommands/features/ClientTimeModifier.java
Packsolite: Tbh this makes no difference in my opinion as the enum already is declared in the same class, but I moved it nontheless for you ^^

Packsolite:
> would it be possible to tell the client that doDaylightCycle is false if the type is LOCK? That way, it wouldn't try to interpolate the sun for rendering every tick.

Actually, the client does not interpolate the skybox (unless i'm blind looking at the code). It is only checking for `doDaylightCycle` when increasing the world time each tick. You can see the sun moving at 20 fps when zooming in or turning down fov 😅

src/main/java/net/earthcomputer/clientcommands/features/ClientTimeModifier.java
```diff
@@ -0,0 +1,43 @@
...
+            this.timeMappingFunction = timeMappingFunction;
+        }
+
+        public long getModifiedTime(long timeOfDay) {
+            return timeMappingFunction.apply(timeOfDay);
+        }
+    }
+
+    private static Type type = Type.NONE;
+    private static long value = 0;
```
Earthcomputer, line 24: Society if Java had Rust-style enums

Earthcomputer (approved):

PR #406, opened 2022-05-08
reflectively register commands

kibibites:
reduced code size idk


Earthcomputer (changes requested):
No thanks, it's not worth the extra dependency, sorry :/

PR #411, opened 2022-05-21
Add the cvar command

xpple:
This allows the player to add custom variables to use in messages and commands. A player may for example use the variable `%base%` so that they don't have to remember the coordinates of their base:
```
The coordinates for my base are %base%
```
This PR has been marked as a draft because using variables in commands currently disrupts the command suggestions. If this is fine for now, I can mark the PR as ready.


src/main/java/net/earthcomputer/clientcommands/command/VarCommand.java
```diff
@@ -0,0 +1,154 @@
...
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import static com.mojang.brigadier.arguments.StringArgumentType.*;
+import static net.fabricmc.fabric.api.client.command.v1.ClientCommandManager.*;
+import static net.minecraft.command.CommandSource.*;
+
+public class VarCommand {
+
+    private static final Pattern VARIABLE_PATTERN = Pattern.compile("[%]([^%]+)[%]");
```
Earthcomputer: This regex could be simplified: `%([^%]+)%`

src/main/java/net/earthcomputer/clientcommands/command/VarCommand.java
```diff
@@ -0,0 +1,154 @@
...
+
+    private static final Logger LOGGER = LogManager.getLogger("clientcommands");
```
Earthcomputer: We're using slf4j now

Earthcomputer (changes requested):

src/main/java/net/earthcomputer/clientcommands/command/VarCommand.java
```diff
@@ -0,0 +1,154 @@
...
+import java.io.IOException;
+import java.nio.file.Path;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import static com.mojang.brigadier.arguments.StringArgumentType.*;
+import static net.fabricmc.fabric.api.client.command.v2.ClientCommandManager.*;
+import static net.minecraft.command.CommandSource.suggestMatching;
```
Earthcomputer: `*` for static imports (or in this case, no static import is also acceptable)

Earthcomputer (approved):

PR #413, opened 2022-05-25
Add the csnake command

xpple:
This PR adds a simple snake game that can be played inside Minecraft. For now this is it, additional features can always be added later.
image


src/main/java/net/earthcomputer/clientcommands/command/SnakeCommand.java
```diff
@@ -0,0 +1,158 @@
...
+
+    private int tickCounter = 10;
+    private Direction direction = Direction.EAST;
+    private Direction lastMoved = Direction.EAST;
+    private final LinkedList<Vec3i> snake = new LinkedList<>();
+    private Vec3i apple;
+
+    SnakeGameScreen() {
+        super(new TranslatableText("snakeGame.title"));
+        this.snake.add(new Vec3i(6, -1184951860, 8));
```
Earthcomputer: What is this magic constant?

src/main/java/net/earthcomputer/clientcommands/command/SnakeCommand.java
```diff
@@ -0,0 +1,158 @@
...
+import static net.fabricmc.fabric.api.client.command.v1.ClientCommandManager.*;
+
+public class SnakeCommand {
+    public static void register(CommandDispatcher<FabricClientCommandSource> dispatcher) {
+        dispatcher.register(literal("csnake")
+                .executes(ctx -> snake(ctx.getSource())));
+    }
+
+    private static int snake(FabricClientCommandSource source) {
+        source.getClient().send(() -> source.getClient().setScreen(new SnakeGameScreen()));
```
Earthcomputer, line 37: This `send` invocation is to prevent the screen from being overwritten by the chat screen closing right? Better add a comment to explain that as it's not obvious.

src/main/java/net/earthcomputer/clientcommands/command/SnakeCommand.java
```diff
@@ -0,0 +1,158 @@
...
+
+    private static final Random random = new Random();
+
+    private static final int MAX_X = 16;
+    private static final int MAX_Z = 16;
+
+    private int tickCounter = 10;
+    private Direction direction = Direction.EAST;
+    private Direction lastMoved = Direction.EAST;
+    private final LinkedList<Vec3i> snake = new LinkedList<>();
```
Earthcomputer: Slightly better than `LinkedList` would be `ArrayDeque`

Earthcomputer: Actually why not `LinkedHashSet`?

Gaming32: Because it's implemented by taking the last segment and adding it back on to the first, a Map doesn't work. I see the contains call, but that's not really comparing enough things to make a HashMap significantly faster anyway.

Earthcomputer (approved):
Now to add a feature to crack snake RNG

ImUrX:
Snake MVP

PR #451, opened 2022-09-29
Remove namehistory subcommand

### LINKED ISSUE #448: Remove name history access
[Mojang is removing the name history API](https://help.minecraft.net/hc/en-us/articles/8969841895693), so `/cplayerinfo namehistory <player>` will not work after September 13, 2022.

xpple:
I didn't choose to make the warning message translatable because I feel like it's more of a debug message than a user feedback message.


Earthcomputer:
Maybe just remove the command entirely if it has no purpose? It can always be re-added later.

xpple:
Yea @haykam821 pointed this out too, good idea.

Earthcomputer (approved):

PR #455, opened 2022-10-27
Add functions and startup commands

Earthcomputer:
Adds two commands: `/cfunction` and `/cstartup`.

A specially named function (based on the server ip / world name) is automatically run on "startup" (server/world join) if present, which by default calls a common startup function. The `/cstartup` command is just a convenience for creating and editing these special functions.

### DIFF
--- a/src/main/java/net/earthcomputer/clientcommands/command/StartupCommand.java
+++ b/src/main/java/net/earthcomputer/clientcommands/command/StartupCommand.java
@@ -1,0 +1,91 @@
+package net.earthcomputer.clientcommands.command;
+
+import com.mojang.brigadier.Command;
+import com.mojang.brigadier.CommandDispatcher;
+import com.mojang.brigadier.exceptions.CommandSyntaxException;
+import com.mojang.brigadier.exceptions.SimpleCommandExceptionType;
+import com.mojang.logging.LogUtils;
+import net.earthcomputer.clientcommands.features.ClientCommandFunctions;
+import net.fabricmc.fabric.api.client.command.v2.FabricClientCommandSource;
+import net.minecraft.text.Text;
+import net.minecraft.util.Util;
+import org.slf4j.Logger;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.StandardOpenOption;
+
+import static com.mojang.brigadier.arguments.StringArgumentType.*;
+import static net.fabricmc.fabric.api.client.command.v2.ClientCommandManager.*;
+
+public class StartupCommand {
+    private static final Logger LOGGER = LogUtils.getLogger();
+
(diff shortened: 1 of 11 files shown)


src/main/java/net/earthcomputer/clientcommands/features/ClientCommandFunctions.java
```diff
@@ -0,0 +1,252 @@
...
+import java.nio.file.InvalidPathException;
+import java.nio.file.Path;
+import java.util.*;
+import java.util.function.IntConsumer;
+import java.util.stream.Stream;
+
+public class ClientCommandFunctions {
+    private static final Logger LOGGER = LogUtils.getLogger();
+
+    private static final Path FUNCTION_DIR = FabricLoader.getInstance().getConfigDir().resolve("clientcommands").resolve("functions");
```
xpple: I suppose this applies to all commands involving file storage, but should we use the static variable `configDir` from `ClientCommands.java`? If not we could make that variable a local variable.

src/main/java/net/earthcomputer/clientcommands/command/StartupCommand.java
```diff
@@ -0,0 +1,91 @@
...
+                .then(literal("edit")
+                    .executes(ctx -> editStartupCommand(true)))));
+    }
+
+    private static int addStartupCommand(FabricClientCommandSource source, boolean global, String command) throws CommandSyntaxException {
+        ensureStartupFileExists(global);
+
+        Path file = global ? ClientCommandFunctions.getGlobalStartupFunction() : ClientCommandFunctions.getLocalStartupFunction();
+        if (file == null) {
+            throw COULD_NOT_CREATE_FILE_EXCEPTION.create();
```
xpple, line 49: No file is being created here, right?

src/main/java/net/earthcomputer/clientcommands/command/StartupCommand.java
```diff
@@ -0,0 +1,91 @@
...
+
+        return Command.SINGLE_SUCCESS;
+    }
+
+    private static int editStartupCommand(boolean global) throws CommandSyntaxException {
+        ensureStartupFileExists(global);
+
+        Path file = global ? ClientCommandFunctions.getGlobalStartupFunction() : ClientCommandFunctions.getLocalStartupFunction();
+        if (file == null) {
+            throw COULD_NOT_CREATE_FILE_EXCEPTION.create();
```
xpple, line 69: Same here.

src/main/java/net/earthcomputer/clientcommands/features/ClientCommandFunctions.java
Earthcomputer: I forgot that existed

src/main/java/net/earthcomputer/clientcommands/command/StartupCommand.java
Earthcomputer: If `file` is null, that means that it was impossible to create the file a few lines above.

PR #429, opened 2022-06-24
whisper using player's mojang provided public key to encrypt

wagyourtail:
this is bad xd

PR #461, opened 2022-11-06
C2C Packets

xpple:
Closes #429. I have implemented a protocol for sending arbitrary encrypted packets through `/w`. New packets can be added by creating a new class and registering it to `CCPacketHandler.java`. The `StringBuf` class is a bit iffy, but it was the most practical solution I could think of. There are still some bugs, hence the draft pull request.

Thank you @wagyourtail for originally implementing this for encrypted messages.


src/main/java/net/earthcomputer/clientcommands/TempRules.java
```diff
@@ -5,6 +5,7 @@
 import net.earthcomputer.clientcommands.features.FishingCracker;
 import net.earthcomputer.clientcommands.features.PlayerRandCracker;
 import net.minecraft.util.StringIdentifiable;
+import net.minecraft.util.annotation.Debug;
```
Earthcomputer: Unused import

src/main/java/net/earthcomputer/clientcommands/c2c/CCNetworkHandler.java
```diff
@@ -0,0 +1,76 @@
...
+    public static boolean justSent = false;
+
+    private static CCNetworkHandler instance;
+
+    private CCNetworkHandler() {
+    }
+
+    public static CCNetworkHandler getInstance() {
+        if (instance == null) {
+            instance = new CCNetworkHandler();
```
Earthcomputer: The instance field probably doesn't need to be lazy (and can therefore be final)

src/main/java/net/earthcomputer/clientcommands/c2c/CCNetworkHandler.java
```diff
@@ -0,0 +1,76 @@
...
+        }
+        return instance;
+    }
+
+    public void sendPacket(CCPacket packet, PlayerListEntry recipient) throws CommandSyntaxException {
+        Integer id = CCPacketHandler.getId(packet.getClass());
+        if (id == null) {
+            return;
```
Earthcomputer, line 38: Maybe log something here?

src/main/java/net/earthcomputer/clientcommands/c2c/CCNetworkHandler.java
```diff
@@ -0,0 +1,76 @@
...
+
+    @Override
+    public void onMessageC2CPacket(MessageC2CPacket packet) {
+        String sender = packet.getSender();
+        String message = packet.getMessage();
+        MutableText prefix = Text.empty();
+        prefix.append(Text.literal("[").formatted(Formatting.DARK_GRAY));
+        prefix.append(Text.literal("/cwe").formatted(Formatting.AQUA));
+        prefix.append(Text.literal("]").formatted(Formatting.DARK_GRAY));
+        Text text = prefix.append(" " + sender + " -> you: " + message).formatted(Formatting.GRAY);
```
Earthcomputer: This text should be translatable

src/main/java/net/earthcomputer/clientcommands/command/WhisperEncryptedCommand.java
```diff
@@ -0,0 +1,53 @@
...
+                .findFirst()
+                .orElseThrow(PLAYER_NOT_FOUND_EXCEPTION::create);
+
+        MessageC2CPacket packet = new MessageC2CPacket(source.getClient().getNetworkHandler().getProfile().getName(), message);
+        CCNetworkHandler.getInstance().sendPacket(packet, recipient);
+        MutableText prefix = Text.empty();
+        prefix.append(Text.literal("[").formatted(Formatting.DARK_GRAY));
+        prefix.append(Text.literal("/cwe").formatted(Formatting.AQUA));
+        prefix.append(Text.literal("]").formatted(Formatting.DARK_GRAY));
+        Text text = prefix.append(" you -> " + recipient.getProfile().getName() + ": " +  message).formatted(Formatting.GRAY);
```
Earthcomputer: This text should be translatable

src/main/java/net/earthcomputer/clientcommands/c2c/CCNetworkHandler.java
```diff
@@ -0,0 +1,76 @@
...
+
+import java.security.PublicKey;
+
+public class CCNetworkHandler implements CCPacketListener {
+
+    private static final SimpleCommandExceptionType MESSAGE_TOO_LONG_EXCEPTION = new SimpleCommandExceptionType(Text.translatable("ccpacket.messageTooLong"));
+    private static final SimpleCommandExceptionType PUBLIC_KEY_NOT_FOUND_EXCEPTION = new SimpleCommandExceptionType(Text.translatable("ccpacket.publicKeyNotFound"));
+    private static final SimpleCommandExceptionType ENCRYPTION_FAILED_EXCEPTION = new SimpleCommandExceptionType(Text.translatable("ccpacket.encryptionFailed"));
+
+    public static boolean justSent = false;
```
Earthcomputer: Is there some less hacky way to do this than a global variable? i.e. another way to figure out whether a message was sent by yourself or not.

wagyourtail: Because of message signing there should be

xpple: I am not sure. The server sends you a `ChatMessageS2CPacket` when you send a message, as well as receive a message. In any case, the "you whisper to ..." is only printed to the chat when you receive this packet, not directly after executing the command. But I don't think there is a way to distinguish between `ChatMessageS2CPacket`s you receive after sending or after receiving a message. In particular not because many servers tend to use `GameMessagePackets`s instead.

wagyourtail: You could also store sent messages in a cache and if it matches...

xpple: Alright, should I use a cache instead @Earthcomputer?

src/main/java/net/earthcomputer/clientcommands/c2c/PacketCacheHelper.java
```diff
@@ -0,0 +1,24 @@
+package net.earthcomputer.clientcommands.c2c;
+
+import com.google.common.cache.CacheBuilder;
+
+import java.time.Duration;
+import java.util.Collections;
+import java.util.Set;
+
+public class PacketCacheHelper {
```
Earthcomputer: Not keen on the name of this class tbh, although I don't immediately have any better suggestions. It's not indicative of its functionality.

src/main/java/net/earthcomputer/clientcommands/c2c/PacketCacheHelper.java
```diff
@@ -0,0 +1,24 @@
...
+
+    private static final Set<String> cache = Collections.newSetFromMap(CacheBuilder.newBuilder().expireAfterWrite(Duration.ofMinutes(1)).<String, Boolean>build().asMap());
+
+    public static boolean contains(String packetString) {
```
Earthcomputer: This should probably be renamed to indicate that it's removing something. Currently the method name indicates it's an immutable query, and might accidentally be used as such.

src/main/java/net/earthcomputer/clientcommands/c2c/PacketCacheHelper.java
xpple: What about `ExpiringPacketCache`?

Earthcomputer: It's more like a class to prevent you getting your own messages, so maybe the name should reflect that?

xpple: `OutgoingPacketCache`, `OutgoingPacketFilter`, ...?

src/main/java/net/earthcomputer/clientcommands/mixin/MixinClientPlayNetworkHandler.java
```diff
@@ -3,26 +3,37 @@
...
 import org.spongepowered.asm.mixin.injection.Inject;
 import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
 
+import java.nio.charset.StandardCharsets;
+import java.security.PrivateKey;
+import java.util.Arrays;
+import java.util.Optional;
+
 @Mixin(ClientPlayNetworkHandler.class)
 public class MixinClientPlayNetworkHandler {
```
Earthcomputer: The modifications to this class are now unnecessary

xpple:
Would it be an idea to distinguish between normal commands and c2c related commands somehow? Perhaps `cc` as prefix? Then the `WhisperEncryptedCommand` would become `CCWhisperCommand` for example and `/cwe` would become `/ccw`.

Earthcomputer:
I don't see the point of that unless we are going to have lots of them. Also what would `cc` stand for? I had more envisaged this mostly being used as a component for other features.

xpple:
On a different note: currently the prefix for C2C packet strings is `CCENC:`. It is therefore very easy for players to fake-send C2C packets by just writing `CCENC:` in their chat messages. Should we try to prevent this by using some arbitrary symbols instead? In particular those that are hard to write in chat manually.

wagyourtail:
I don't think that's really necessary because it won't cause any adverse effects to the client... the decrypt should fail and leave as is

xpple:
> I don't think that's really necessary because it won't cause any adverse effects to the client... the decrypt should fail and leave as is

This is the current response to malformed C2C packets:
image
If this is fine I'll just leave it as is.

RealIndrit:
> > I don't think that's really necessary because it won't cause any adverse effects to the client... the decrypt should fail and leave as is
> 
> This is the current response to malformed C2C packets: image If this is fine I'll just leave it as is.

I think this looks fine, I personally see no need to try "prohibit" fake CCENC packets, as it will just fail and show error message and the user who sent the malformed packet anyways. 👍

Earthcomputer (approved):

PR #462, opened 2022-11-08
Add /cpos command to convert between coordinates of different dimensions

magicus:
Hi,

Sorry for not opening a feature request issue beforehand, but it felt simpler to just write the code, and have a discussion afterwards, if needed.

This patch adds a `/cnether` command, which converts between Overworld and Nether coordinates. I've found myself far too many times lookup my current coordinate and type `/ccalc <current x> / 8` etc.

`/cnether` can be used without arguments, in which case your current position is used, or it can take a Vanilla position argument. If you are in the Overworld when the command is run, the corresponding Nether coordinates are shown, and conversely, if you are in the Nether, then the corresponding Overworld coordinates are shown. 

For clarity, the message includes your current coordinates as well, and a clear specification of which world they relate to. It is a bit verbose, but I think it is better to be crystal clear when talking about coordinates in different worlds, than too brief and possibly not clear enough.

An example usage, with the player standing at (-21, 68, -9) in the Nether, would be:
```
/cnether ~ ~ ~
(-21, 68, -9) in the Nether is (-168, 68, -72) in the Overworld
```

The latter set of coordinates is linked to a `/clook` command, which perhaps might be useful once on the other side, to find the location where to build your portal. (It will of course point incorrectly in the same world as you issued the command.)

I've found this addition to be very helpful, and I hope it can become accepted to be part of clientcommands.


haykam821:
Could this command be generalized to allow conversions between arbitrary dimensions?

magicus:
@haykam821 I'm trying to understand what you mean. Is your request that you should be able to be in the Nether, and ask "what would 123 45 67 in the overworld be in the nether"?

My initial iterations of this patch actually had something like that, in that it printed two lines, something like this:
```
/cnether ~ ~ ~
(-21, 68, -9) in the Nether is (-168, 68, -72) in the Overworld
If you were in the Overworld, (-21, 68, -9) would have been (-3, 68, -1) in the Nether
```

I found it more annoying and confusing than helpful, so I removed it.

magicus:
There are other possible approaches, of course. I can think of two: having an optional `dimension` argument after the position, or creating two commants, `/cnether` and `/coverworld`. 

Both have usability problems, though. The first has the problem: which dimension are you specifying, the source or the target dimension? Neither seems obviously correct, and I guess most users will be forced to pick one at random and then see if it worked or not.

The second suffers from a somewhat silly-sounding name ("cover world"?), and what seems like unnecessary command proliferation. 

If you have any more and better ideas, I'm all ears.

xpple:
Thanks for your pull request! I think haykam is referring to not hardcoding the number `8`, but rather using `DimensionType#getCoordinateScaleFactor`. You can make a custom argument that makes it explicit what the source dimension and what the target dimension is.

haykam821:
Yes, the coordinate scale for both the current dimension and the Nether should be used by default. The dimensions can be optional, but the target dimension or both dimensions should be specifiable by a user if needed.

magicus:
I can of course add that. I'm just curious what the use case would be. Is it for custom, modded dimensions? (I have not really played with those, just know that they exist.) For Vanilla, there seem to be no reason to convert anything but between the Overworld and the Nether. 

My main worry is that it will complicate the basic premise of the command, a fast and simple conversion between Overworld and Nether coordinates, which just "does what you mean".

haykam821:
This could be for custom dimensions or just a custom coordinate scale. The usages should look like this:

- `cnether` - Get coordinates in `minecraft:the_nether` when traveling from player's position in `minecraft:overworld` (or vice versa)
- `cnether <pos>` - With specific coordinates
- `cnether <pos> <to>` - When going from the current dimension to the dimension specified in the `to` argument
- `cnether <pos> <from> <to>` - When going from the dimension specified in the `from` argument to the dimension specified in the `to` argument

src/main/java/net/earthcomputer/clientcommands/command/NetherCommand.java
```diff
@@ -0,0 +1,54 @@
...
+        } else {
+            targetPos = new BlockPos(pos.getX()/8, pos.getY(), pos.getZ()/8);
+            sourceWorld = I18n.translate("commands.cnether.world.overworld");
+            targetWorld = I18n.translate("commands.cnether.world.nether");
+        }
+
+        source.sendFeedback(getCoordsTextComponent(pos)
+                .append(Text.translatable("commands.cnether.coords.left", sourceWorld, targetWorld))
+                .append(getLookCoordsTextComponent(targetPos))
+                .append(Text.translatable("commands.cnether.coords.right", sourceWorld, targetWorld)));
```
haykam821: Prefer `TranslatableText` arguments over appending text together.

wagyourtail:
the to/from can be before pos in the later and still work, since they're different types the pos should be prefered, at least as long as it's registered first...
so:
`cnether`, `cnether <pos>`, `cnether <to> <pos>`, `cnether <to> <from> <pos>`

magicus:
With these suggested changes, it probably makes more sense to rename the command `/ccoords`..? Or perhaps `/cpos`, it sounds less clunky.

xpple:
Yeah, perhaps `/cpos [in|from <dimension>] [in|from <dimension>]`. Then the command without any arguments would use the player's current dimension, and convert it from/to the nether dimension. When `in <dimension>` is specified, that dimension is used as target dimension (read it as "[my current] `/cpos in <dimension>`"). When `from <dimension>` is specified, that dimension will be used as the source dimension. Then the premise of the command - being short and quick - isn't complicated, but it still has all the functionality you'd expect. Optionally a position could be specified at the very end of the command.

magicus:
I'll experiment a bit with the different syntax suggestions, and see what seems to work fine. I do think however that you need to be able to enter the position without any hassle, since one of the benefits of the original command was that you could stand next to your nether portal, look at one of the blocks at the base of the portal and press tab, and get the coordinates for that block. I don't want to see this functionality lost.

Or, if you are open to more commands, there might be room for both a simplified `/cnether` which is limited to converting to/from nether coordinates (more or less, the current implementation) and a more advanced `/cpos` which takes a source and target dimension and an optional position, as you suggest. I kind of liked the intuitivity and readability of just typing `/cnether`, so I think that would be a good compromise between keeping this simplicity with having the full power that you are requesting.

magicus:
Getting back to trying to get this to work. It was a bit more tricky than I expected, especially the part about reading the scale factor from the Vanilla information. 

It turns out that the scale information resides in the `DimensionType`. There is even a helpful `getCoordinateScaleFactor(DimensionType fromDimension, DimensionType toDimension)` that I'd like to use. And I also found a nice `dev.xpple.clientarguments.arguments.CDimensionArgumentType` which allowed me to add  DimensionArgumentType arguments to the `/cpos` command. Nice!

However... 😞  

`CDimensionArgumentType` only gives me a `RegistryKey<World>`, and try as I might, I fail to find a way to get an associated `DimensionType` from this. From an actual `World` instance, I can call `getDimension()`. But there seems to be no way to go from a `RegistryKey<World>` to a `World`. As far as I can tell this separation is by design. The command only has access to the current `ClientWorld`, and the list of all `World`s is created in `MinecraftServer#createWorlds`, so technically the client would not have access to them. It kind of makes sense, but it also kind of extremely annoying. :(

There is a `BuiltinRegistries.DIMENSION_TYPE`, which will allow me to go from a `RegistryKey<DimensionType>` to a `DimensionType`. So one way forward would be to create a parallel to `CDimensionArgumentType` that returns `RegistryKey<DimensionType>` instead of `RegistryKey<World>`. However, this registry also contains `DimensionTypes.OVERWORLD_CAVES` so I'd have to either filter that out, or hard-code the three normal dimension types, or accept that "overworld_caves" will appear as a dimension argument suggestion. (Neither option sounds very attractive...)

I might also be able to use the fact that both the `World` and `DimensionType` registry keys share the same identifier string ("the_nether" etc). But I don't really like this solution either; it would rely more on "coincidence" than guaranteed relations.

I would appreciate any suggestions on how to proceed. In the meantime, I've pushed my updates that implements (more or less) the syntax suggested by @xpple above, but still hardcoding the scale factor 8.

(My opinion is that this is as good as it gets. If my understanding is correct, we might be using the client to play on a server which has a non-traditional scale factor, but if that is the case, we will not be able to get hold of this scale factor anyway.)

haykam821:
You can access a value using a registry key through the dynamic registry manager. You should have access to one as it is synced to the client.

magicus:
The new `/cpos` argument works like this:
`/cpos` with no argument assumes the player's current position, the current dimension as the source dimension, and the "opposite" dimension as the target. (The "opposite" dimension is defined as the Nether for all dimensions except the Nether, and the Overworld for the Nether).

If the user specifies `from <dimension>` that will be used as the source dimension, and the target will be the "opposite". If the user specifies `to <dimension>`, that will be used as the target dimension, and the source will be the opposite. If the user specifies both `from` and `to`, well, then both these are used as source and target.

Due to a bit of trickery and complexity in the `register` method, the user can specify zero, one or both of `from` and `to`, and in any order they like. The position is always optional, and is always placed last, if given.

I think the GUI aspects of the command worked out quite nicely. I still miss a bit the kind of clear connection between the name and result of `/cnether`, but I guess this is okay as well.

magicus:
@haykam821 As far as I can tell, the `World`s are not registered with the registry manager. 

How should I access them, do you mean? 

`Registry<World> reg = DynamicRegistryManager.BUILTIN.get().get(Registry.WORLD_KEY);` only gives me an exception: "Missing registry: ResourceKey[minecraft:root / minecraft:dimension]"

magicus:
As I said, `DimensionType`s are available from the registry manager, but the actual `World`s are not.

src/main/java/net/earthcomputer/clientcommands/command/NetherCommand.java
magicus: Well, this was based around the existing `getLookCoordsTextComponent()` will will return properly formatted coordinates, with a onClick event to send a `/clook` command. That means it *must* be a separate Text node, or the withClickEvent would apply to the entire string. So I needed at least three separate Text blocks -- the "look coords" and the block prior and after.

Due to the symmetry of the response, I thought it made more sense to keep this symmetry in how the feedback string was built up. We're not talking performance critical code here, so I opted to strive for good code readability instead.

haykam821:
`DynamicRegistryManager.BUILTIN` is not the correct dynamic registry manager. You should access the client's registry manager, which I believe can be accessed through `source.getClient().getRegistryManager()`.

magicus:
You will have to be more specific. `source.getClient()` returns `MinecraftClient`. There is no `getRegistryManager()` in that class. I have searched the source code for both "registry" and "manager" to try and find anything close to that, but to no avail.

I think you are just mistaken. I have traced the code when creating a `World`. As far as I can see, it is never stored in a registry.

magicus:
To be super clear: the created worlds are only stored in `MinecraftServer` in the field `private final Map<RegistryKey<World>, ServerWorld> worlds`. From the `FabricClientCommandSource` I have only access to the current `ClientWorld`. I think what you are asking for is essentially impossible.

src/main/java/net/earthcomputer/clientcommands/command/PosCommand.java
```diff
@@ -0,0 +1,136 @@
...
+                            .then(argument("pos", blockPos())
+                                    .executes(ctx -> convertCoords(ctx.getSource(),
+                                            getCBlockPos(ctx, "pos"), getCDimensionArgument(ctx, "sourceDimension"), getCDimensionArgument(ctx, "targetDimension"))))
+                            .executes(ctx -> convertCoords(ctx.getSource(),
+                                    ctx.getSource().getPlayer().getBlockPos(), getCDimensionArgument(ctx, "sourceDimension"), getCDimensionArgument(ctx, "targetDimension")))))
+                    .executes(ctx -> convertCoords(ctx.getSource(),
+                            ctx.getSource().getPlayer().getBlockPos(), getCDimensionArgument(ctx, "sourceDimension"), null))))
+            .executes(ctx -> convertCoords(ctx.getSource(),
+                    ctx.getSource().getPlayer().getBlockPos(), null, null)));
+    }
```
xpple, line 53: I'd prefer making a custom `ArgumentType` for this command.

src/main/java/net/earthcomputer/clientcommands/command/PosCommand.java
```diff
@@ -0,0 +1,136 @@
...
+import net.minecraft.util.registry.RegistryKey;
+import net.minecraft.world.World;
+
+import static dev.xpple.clientarguments.arguments.CBlockPosArgumentType.blockPos;
+import static dev.xpple.clientarguments.arguments.CBlockPosArgumentType.getCBlockPos;
+import static dev.xpple.clientarguments.arguments.CDimensionArgumentType.dimension;
+import static dev.xpple.clientarguments.arguments.CDimensionArgumentType.getCDimensionArgument;
+import static net.earthcomputer.clientcommands.command.ClientCommandHelper.getLookCoordsTextComponent;
+import static net.fabricmc.fabric.api.client.command.v2.ClientCommandManager.argument;
+import static net.fabricmc.fabric.api.client.command.v2.ClientCommandManager.literal;
```
xpple: Please use wildcard imports for static imports.

xpple (changes requested):
Looking good!

Also, there is a `getRegistryManager` method available from the command source.

Gaming32:
You could also add support for converting between other dimensions' coordinates. There's a coordinate scale in the dimension JSON that is relative to the overworld.

magicus:
@Gaming32 What do you mean? What is that "dimension JSON"?

magicus:
> Also, there is a `getRegistryManager` method available from the command source.

Oh, that's good to know. It still does not have anything registered for `Registry.WORLD_KEY` though.

src/main/java/net/earthcomputer/clientcommands/command/PosCommand.java
magicus: Okaaaay... trying to understand what you want. Do you want the coupled literal prefix and the dimension type to be a separate `ArgumentType`?

The argument chain is rather complex, I agree, but that is to make sure that a command like `/cpos from the_nether from overworld` can not be entered -- if you go down the `from` path, the second argument can only take a `to`, and vice versa. 

So while I *could* make e.g. a `SelectDimensionArgumentType`, it would lose the ability to prohibit this.

Or do you mean that I should shove the entire thing into a `CPosArgumentType`? I fail to see how that would be an improvement; the `ArgumentType` is intended for "basic" building blocks, and I would not have any easy access to this kind of chaining of methods that describe the syntax. Instead I'd have to resort to more basic parsing approaches.

To be honest, I've already spent much more time on this than I anticipated with my original simple `/cnether` command. At this point I'd rather drop the PR than rewrite this complex chain of arguments as a parser.

xpple: A custom argument yields more flexibility and control over the command syntax than using the command builder. You could implement an argument type much like `TranslationQueryArgumentType`, but with strict conditions to prevent duplicate clauses.

> To be honest, I've already spent much more time on this than I anticipated with my original simple `/cnether` command. At this point I'd rather drop the PR than rewrite this complex chain of arguments as a parser.

That's understandable. If @Earthcomputer thinks the command is fine as is, I have no problem with it being merged. In any case thanks for your efforts!

Earthcomputer: Yeah I don't see this chain to be a problem. If it were to become more complex in the future for some reason then that would be the case for changing it to an argument type, but I don't see how that's possible in this case.

Earthcomputer (approved):

haykam821:
Looking into it better, you're right about how worlds are stored. They don't seem to be part of dynamic registry manager as seen with my test code:

```java
DynamicRegistryManager manager = source.getWorld().getRegistryManager();

for (DynamicRegistryManager.Entry<?> registry : manager.streamAllRegistries().toList()) {
	Identifier id = registry.key().getValue();
	source.sendFeedback(Text.literal(id + ""));
}

Registry<World> worlds = manager.get(Registry.WORLD_KEY);

for (World world : worlds) {
	Identifier id = world.getRegistryKey().getValue();
	double scale = world.getDimension().coordinateScale();

	source.sendFeedback(Text.literal(id + ": " + scale));
}
```

At the very least, the dimension type of the current world could be accessed. While not necessary, the coordinate scales of previously seen worlds could be cached as well. Since the fallback requires assumption, could a warning be printed instead?

src/main/java/net/earthcomputer/clientcommands/command/NetherCommand.java
haykam821: I don't see what you mean. Wouldn't both solutions would allow separate click events for the coordinates? This can be seen in the death message for a bad respawn point, for example:

```java
MutableText link = Texts.bracketed(Text.translatable("death.attack.badRespawnPoint.link")).styled(style -> {
	return style
		.withClickEvent(new ClickEvent(ClickEvent.Action.OPEN_URL, "https://bugs.mojang.com/browse/MCPE-28723"))
		.withHoverEvent(new HoverEvent(HoverEvent.Action.SHOW_TEXT, Text.literal("MCPE-28723")))
});

return Text.translatable("death.attack.badRespawnPoint.message", entity.getDisplayName(), link);
```

 My suggestion was to use a single `commands.cpos.coords` rather than the separate `commands.cpos.coords.left` and `commands.cpos.coords.right`.

magicus:
@xpple Are you happy with the change now?

wagyourtail:
> Looking into it better, you're right about how worlds are stored. They don't seem to be part of dynamic registry manager as seen with my test code:...

is there not a way to work directly with DIMENSION_TYPE

haykam821:
The dimension type could be used as the argument, yes. It could potentially cause confusion especially with multiple dimensions using one dimension type, but wouldn't have the issue of the client not having the associated dimension type for an arbitrary dimension. Do you think it's better to use the dimension type as the argument?

Gaming32:
Isn't there a dimension argument type? You know, what's used by `/execute in `.

EDIT: `net.minecraft.command.argument.DimensionArgumentType` does exist in Vanilla.

Earthcomputer:
Yes and they're using the clientarguments version of that. You can't get the world of another dimension on the client because it doesn't exist.

src/main/java/net/earthcomputer/clientcommands/command/NetherCommand.java
Earthcomputer: That's a valid issue however I reckon we should solve it after the PR has been merged, because it's a problem around the code base.

Earthcomputer:
I reckon this is fine to merge now. Not being able to use `coordinateScale` sucks but oh well, nothing we can do.

PR #464, opened 2022-11-13
Use ExtendedMarkdownArgument for /cwe

xpple:
I rewrote `FormattedTextArgumentType`. The previous implementation was too verbose and moreover not working as one might have expected it to. This new syntax is less error prone and probably more intuitive to work with. The availability of some text _styles_ have been dropped, as firstly the previous implementation was incomplete and secondly because this argument type is meant to provide `Formatting`ed text, not `Style`d text. Aside from what's available in the `Formatting` class, I did decide to add hex colour support through the `#` prefix with lenient hex parsing. Do note that `Style`d text is still available using the `/ctellraw` command.


Earthcomputer:
Hmm, could we leave in some of the old syntax as an alternative to the new syntax, so that those of us who haven't memorized formatting codes can still use it?

xpple:
I think that will get a bit confusing. What about making this into a separate class? Something like `LegacyFormattedTextArgumentType`. Then the `/cnote` command would just remain as is. My intentions for this class were to use this formatting syntax to be able to write formatted messages over C2C packets in a very compact and non distracting way. Being spammed with formatting suggestions all the time would get annoying if one just wanted to send a simple message.

xpple:
I have updated the PR. Instead of replacing the existing `StyledComponentArgument` (which I still think sucks), I have now separated this new argument type. I have furthermore added it to `/cwe`. In the C2C packet the string is not encoded as formatted text usually is, which is very lengthy. Instead the raw argument string is sent, with the use of the `WithStringArgument`.
image
image

Earthcomputer (changes requested):
https://github.com/Earthcomputer/clientcommands/pull/464#issuecomment-1320938132 you still haven't addressed this comment, but I want to look into rewriting the formatted component argument type to look more like markdown

xpple:
Well I addressed that by creating a separate one as opposed to rewriting the current one. And to be honest most clientcommands users will already be familiar with this `&` formatting due to their common usage in Spigot servers. And if not, this syntax is nondistracting so it will not get in the way if one wants to send a simple message and also (surprisingly perhaps) easy to learn. I know the syntax is not established in the modded server community but it is definitely in the Spigot server community, which most of the clientcommands users (to my experience) are in. I would say it is a decent syntax because it is good at its purpose.

Markdown is also an option, but I do not think a great one. For one it has no notion of colouring and we would only be able to support a perhaps arbitrary to the user subset of it, unless of course we implement our own client formatting, which I find to be beyond the scope of clientcommands c.q. this PR. I do know of the existence of certain Markdown specs that do support coloured text, but they are not standard at all, so the user would have to learn a new syntax which we probably do not want to introduce.

Lastly, as with all choices, it could be made configurable.

Earthcomputer (changes requested):
Wrote a new argument type in #673 which is now merged, so this needs to be modified to use that

xpple:
I updated it.

Earthcomputer (approved):

PR #469, opened 2022-12-04
Create canned_responses.yml

wagyourtail:
idk if this will work. hard to test...


wagyourtail:
wait. does https://github.com/Earthcomputer/clientcommands/blob/fabric/.github/earthbot.yml still work...

Earthcomputer:
It should do

PR #479, opened 2022-12-19
implement over-engineered diceroll

wagyourtail:
uses a modified diffie-hellman key exchange to prevent cheating


xpple:
Sorry if this is a strange request but could you perhaps extract the C2C packet changes to a separate PR?

wagyourtail:
I can't completely remove them from this one or it'll break this one. but I can have them as a seperate branch that this one is based uppon

xpple:
Sure, thanks!

Starmania:
Would it be useful to use 2048 bit keys? [A wikipedia article](https://en.wikipedia.org/wiki/Diffie–Hellman_key_exchange#Practical_attacks_on_Internet_traffic)

wagyourtail:
> Would it be useful to use 2048 bit keys? [A wikipedia article](https://en.wikipedia.org/wiki/Diffie%E2%80%93Hellman_key_exchange#Practical_attacks_on_Internet_traffic)

yes, but they don't fit in a minecraft chat message, xd and for our purposes I don't think that'll be an issue

src/main/java/net/earthcomputer/clientcommands/c2c/C2CPacket.java
```diff
@@ -1,9 +1,10 @@
 package net.earthcomputer.clientcommands.c2c;
 
+import com.mojang.brigadier.exceptions.CommandSyntaxException;
 import net.minecraft.network.PacketByteBuf;
 
 public interface C2CPacket {
     void write(PacketByteBuf buf);
 
-    void apply(CCPacketListener listener);
+    void apply(CCPacketListener listener) throws CommandSyntaxException;
```
xpple, line 9: I feel like this is out of the scope of command syntax exceptions. Perhaps we should create our own exception class or just `return` instead of throwing an exception.

src/main/java/net/earthcomputer/clientcommands/c2c/packets/DiceRollC2CPackets.java
```diff
@@ -0,0 +1,102 @@
...
+        public void apply(CCPacketListener listener) {
+            listener.onCoinflipResultC2CPacket(this);
+        }
+    }
+
+    public static void register() {
+        CCPacketHandler.register(DiceRollInitC2CPacket.class, DiceRollInitC2CPacket::new);
+        CCPacketHandler.register(DiceRollAcceptedC2CPacket.class, DiceRollAcceptedC2CPacket::new);
+        CCPacketHandler.register(DiceRollResultC2CPacket.class, DiceRollResultC2CPacket::new);
+    }
```
xpple, line 101: Not sure if it's necessary to register them here. I don't know why `CCPacketHandler#register` wasn't private in the first place.

src/main/java/net/earthcomputer/clientcommands/c2c/CCNetworkHandler.java
```diff
@@ -4,7 +4,10 @@
 import com.mojang.brigadier.exceptions.DynamicCommandExceptionType;
 import com.mojang.brigadier.exceptions.SimpleCommandExceptionType;
 import com.mojang.logging.LogUtils;
+import io.netty.buffer.Unpooled;
```
xpple, line 7: Unused import.
```suggestion
```

src/main/java/net/earthcomputer/clientcommands/command/DiceRollCommand.java
```diff
@@ -0,0 +1,279 @@
...
+import org.slf4j.Logger;
+
+import java.math.BigInteger;
+import java.nio.ByteBuffer;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.security.SecureRandom;
+import java.util.*;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
```
xpple, line 27: Unused import.
```suggestion
```

src/main/java/net/earthcomputer/clientcommands/command/DiceRollCommand.java
```diff
@@ -0,0 +1,279 @@
...
+                .executes(ctx -> diceroll(ctx.getSource(), getCProfileArgument(ctx, "player"), 2))));
+
+        dispatcher.register(literal("cdiceroll")
+                .executes(ctx -> localDiceroll(ctx.getSource(), 6))
+                .then(argument("sides", integer(2, 100))
+                        .executes(ctx -> localDiceroll(ctx.getSource(), getInteger(ctx, "sides")))
+                        .then(argument("player", gameProfile())
+                            .executes(ctx -> diceroll(ctx.getSource(), getCProfileArgument(ctx, "player"), getInteger(ctx, "sides")))))
+                .then(argument("player", gameProfile())
+                        .executes(ctx -> diceroll(ctx.getSource(), getCProfileArgument(ctx, "player"), 6))));
```
xpple, line 73: Inconsistent indentation.

src/main/resources/assets/clientcommands/lang/en_us.json
```diff
@@ -110,6 +110,17 @@
...
+  "commands.diceroll.diceroll": "dice roll (%s side)",
+  "commands.diceroll.coinflip": "coin flip",
+  "commands.diceroll.sent": "Sent %s request to %s",
+  "commands.diceroll.received": "Received %s request from %s",
+  "commands.diceroll.value": "%s result from %s: %s",
+  "commands.diceroll.heads": "Heads",
+  "commands.diceroll.tails": "Tails",
+  "commands.diceroll.cheater": "I think %s is cheating, they got a different value.",
+  "commands.diceroll.timeout": "Timeout waiting for %s's response to %s request",
+  "commands.diceroll.waitingForResponse": "Waiting for %s's response already",
```
xpple, line 122: It should be `cdiceroll`. Also I think some of these shouldn't be a subkey of `commands`, but rather of `ccpacket`.

src/main/java/net/earthcomputer/clientcommands/command/DiceRollCommand.java
```diff
@@ -0,0 +1,279 @@
...
+        }
+        return Command.SINGLE_SUCCESS;
+    }
+
+    public static byte[] toSHA1(byte[] convertme) {
+        MessageDigest md;
+        try {
+            md = MessageDigest.getInstance("SHA-1");
+        }
+        catch(NoSuchAlgorithmException e) {
```
xpple, line 91:
```suggestion
        catch (NoSuchAlgorithmException e) {
```

src/main/java/net/earthcomputer/clientcommands/command/DiceRollCommand.java
```diff
@@ -0,0 +1,279 @@
...
+                .findFirst()
+                .orElseThrow(PLAYER_NOT_FOUND_EXCEPTION::create);
+
+        BigInteger a = playerAB.get(packet.sender);
+        BigInteger B = packet.AB;
+
+        // check if hash matches
+        if (!Arrays.equals(opponentHash.get(packet.sender), toSHA1(B.toByteArray()))) {
+            System.out.println("expected: " + byteArrayToHexString(opponentHash.get(packet.sender)));
+            System.out.println("actual: " + byteArrayToHexString(toSHA1(B.toByteArray())));
```
xpple, line 221: This should probably be removed.

src/main/java/net/earthcomputer/clientcommands/command/DiceRollCommand.java
```diff
@@ -0,0 +1,279 @@
...
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
+
+import static com.mojang.brigadier.arguments.IntegerArgumentType.*;
+import static dev.xpple.clientarguments.arguments.CGameProfileArgumentType.*;
+import static net.fabricmc.fabric.api.client.command.v2.ClientCommandManager.*;
+
+public class DiceRollCommand {
+    private static final Logger LOGGER = LogUtils.getLogger();
+    private static final SimpleCommandExceptionType PLAYER_NOT_FOUND_EXCEPTION = new SimpleCommandExceptionType(Text.translatable("commands.cwe.playerNotFound"));
```
xpple, line 35: Perhaps it's an idea to add a few common exceptions for c2c related stuff to a certain class, I don't like the `commands.cwe` translation key being used here.

xpple (changes requested):
I don't understand much of the technical stuff so these are mostly style related changes.

RealRTTV:
might i ask: how does this work, can't i just generate these values on my client over and over until i get the one i want, then send it?

Semisol:
A simpler solution for fairness could be implemented:
- Initiator generates `s_i`, sends `H(s_i)` (commit)
- Receiver saves `H(s_i)`, generates `s_r`, sends `H(s_r)` (commit)
- Initiator saves `H(s_r)`, sends `s_i` (reveal)
- Receiver verifies `H(s_i received) = H(s_i) saved`, calculates xor `s_i ^ s_r = s`, sends `s_r`, shows dice roll result (reveal)
- Initiator verifies `H(s_r received) = H(s_r) saved`, calculates xor `s_i ^ s_r = s`, shows dice roll result

This is secure, as:
- the committed value cannot be changed after the fact, and no values are revealed before commit is complete, so you cannot bias your value to get an expected xor output
- xoring two values that are not correlated can only increase entropy (correlated being generated from the same source as the other value, or generated with knowledge of the other value to intentionally bias the xor)
- there is an attack where receiver can withhold revealing, but this should result in a big warning

PR #487, opened 2022-12-31
Add support for French

Starmania:
I skip some trads because the original is better than one in French, and I don't know how to translate some things.

I used **Json with comments**, and the construction passes, but I don't know if this is the right way to do it.

### DIFF
--- a/src/main/java/net/earthcomputer/clientcommands/render/RenderQueue.java
+++ b/src/main/java/net/earthcomputer/clientcommands/render/RenderQueue.java
@@ -1,6 +1,5 @@
 package net.earthcomputer.clientcommands.render;
 
-import com.mojang.blaze3d.systems.RenderSystem;
 import net.minecraft.client.render.RenderLayer;
 import net.minecraft.client.render.VertexConsumer;
 import net.minecraft.client.render.VertexFormat;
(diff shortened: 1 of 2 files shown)


xpple:
Thank you for your efforts! Could the comments be removed? I don't think they are very helpful, and more importantly I don't think the JSON parser Minecraft uses allows comments. Also, why are some lines translated but also commented out?

Starmania:
I try in-game and the parser of Minecraft skip comments, so this is fine.

Starmania:
And for some lines translated but commented, it's because the translation doesn't make sense (but I keep it for whoever wants to do it) or because I had no idea of a good translation.

Starmania:
So, what I do, I remove comments yes or no ?

Earthcomputer:
Yes please remove the comments. along with any translation that is still in English (I can see a few). The game will default to the English translation if there is no French one found.

Starmania:
Ok, I edited files, they should be good

Earthcomputer:
Any chance you could finish off the missing translations? If you want clarification on what they mean then just ask. If not then that's fine.

Starmania:
Sorry, but I have no idea for a good traduction (and I'm native...)

PR #498, opened 2023-01-20
Big code style refactor

xpple:
You'd think I'd have something better to do with exams coming next week, but no.


src/main/java/net/earthcomputer/clientcommands/command/AliasCommand.java
```diff
@@ -116,14 +118,14 @@ private static int addAlias(FabricClientCommandSource source, String key, String
         if (aliasMap.containsKey(key)) {
             throw ALIAS_EXISTS_EXCEPTION.create(key);
         }
-        if (ClientCommandInternals.getActiveDispatcher().getRoot().getChildren().stream().map(CommandNode::getName).anyMatch(literal -> literal.equals(key))) {
+        if (ClientCommandManager.getActiveDispatcher().getRoot().getChildren().stream().map(CommandNode::getName).anyMatch(literal -> literal.equals(key))) {
```
Earthcomputer, line 114: How did this not break on Quilt?

src/main/java/net/earthcomputer/clientcommands/features/FishingCracker.java
```diff
@@ -277,7 +277,7 @@ public void body() {
                     @Override
                     public void onCompleted() {
                         synchronized (STATE_LOCK){
-                            if(throwFishingBobber()){
+                            if (throwFishingBobber()){
```
Earthcomputer: While you're at it, space before the { as well

src/main/java/net/earthcomputer/clientcommands/features/PlayerRandCracker.java
```diff
@@ -424,8 +435,7 @@ public enum Type {
             NOT_ENOUGH_ITEMS(false, "playerManip.notEnoughItems"),
             NOT_POSSIBLE(false, "playerManip.throwError"),
             UNKNOWN_SEED(false, "playerManip.uncracked"),
-            SUCCESS(true, null),
-            ;
+            SUCCESS(true, null);
```
Earthcomputer: The reason it was like how it was before is because it makes the diff cleaner when you add more types

src/main/java/net/earthcomputer/clientcommands/features/PlayerRandCracker.java
```diff
@@ -454,8 +464,7 @@ public static enum CrackState implements StringIdentifiable {
         CRACKING("cracking"),
         EATING("eating"),
         MANIPULATING_ENCHANTMENTS("manipulating_enchantments"),
-        WAITING_DUMMY_ENCHANT("waiting_dummy_enchant", true),
-        ;
+        WAITING_DUMMY_ENCHANT("waiting_dummy_enchant", true);
```
Earthcomputer: Same here

src/main/java/net/earthcomputer/clientcommands/command/AliasCommand.java
xpple: What do you mean?

src/main/java/net/earthcomputer/clientcommands/features/PlayerRandCracker.java
xpple: I see, do you want me to revert the change?

Earthcomputer: Yes please

src/main/java/net/earthcomputer/clientcommands/command/AliasCommand.java
Earthcomputer: Using internals should have broken on Quilt because Quilt only implements the API part of Fabric. Anyway, doesn't matter

xpple: That's what this code is for
```java
try {
    Class.forName("net.fabricmc.fabric.impl.command.client.ClientCommandInternals");
} catch (ClassNotFoundException e) {
    LOGGER.warn("Could not find ClientCommandInternals, /calias command not available");
    return;
}
```

Earthcomputer: Looks like that code can be removed now.

xpple:
I removed the throws clause because vanilla also doesn't have this and moreover I don't think `CommandSyntaxException`s are suitable here. Whenever you need to catch a `CommandSyntaxException`, it's probably being misused.

xpple:
Lmk if I should revert this change. It's only a slight simplification.

TgZ39:
put it in the main branch, this guy has worked hard to make this pull request

Earthcomputer (approved):
Looks good, just needs conflicts resolving

Earthcomputer (approved):

TgZ39:
pog

PR #513, opened 2023-03-27
Add the cwaypoint command

### LINKED ISSUE #510: Waypoints
A command that allows the player to set waypoints. The original idea is by @ImUrX.

xpple:
Closes #510. Waypoint labels are displayed at the top of the screen in the direction of their location. If two waypoint labels would overlap, they are stacked above each other. Every tick I am creating a list of the $x$-offsets of the waypoint labels calculated against the waypoint location, and player's position and orientation. You could also precompute a data structure that lets you construct a sorted list from the player's position and orientation, but complexity-wise both are $\Theta(n\log(n))$ work, where $n$ is the number of waypoints.


src/main/java/net/earthcomputer/clientcommands/command/WaypointCommand.java
```diff
@@ -0,0 +1,166 @@
...
+            .then(literal("list")
+                .executes(ctx -> list(ctx.getSource()))
+                .then(argument("current", bool())
+                    .executes(ctx -> list(ctx.getSource(), getBool(ctx, "current"))))));
+    }
+
+    private static String getWorldIdentifier(FabricClientCommandSource source) {
+        String worldIdentifier;
+        if (source.getClient().isIntegratedServerRunning()) {
+            worldIdentifier = source.getClient().getServer().getSaveProperties().getLevelName();
```
haykam821: Would this make waypoints disappear from a world if its name is changed?

xpple: That's a good thing to keep in mind for when I actually implement saving to a file.

xpple:
The only thing that remains is actually rendering the waypoints. I'll leave that to someone else as I don't have any experience with rendering stuff like this.

src/main/java/net/earthcomputer/clientcommands/command/WaypointCommand.java
xpple: How do you suggest we do this?

src/main/java/net/earthcomputer/clientcommands/ClientCommands.java
```diff
@@ -91,6 +93,8 @@ public void onInitializeClient() {
         FishingCracker.registerEvents();
         PlayerRandCracker.registerEvents();
         ServerBrandManager.registerEvents();
+        HudRenderCallback.EVENT.register(WaypointCommand::renderWaypointLabels);
+        WorldRenderEvents.AFTER_ENTITIES.register(WaypointCommand::renderWaypointBoxes);
```
Earthcomputer: There could be a `WaypointCommand.registerEvents()` method, just to keep this file a bit cleaner

src/main/java/net/earthcomputer/clientcommands/command/WaypointCommand.java
```diff
@@ -0,0 +1,413 @@
...
+            worldWaypoints.forEach((name, waypoint) -> source.sendFeedback(Component.translatable("commands.cwaypoint.list", name, formatCoordinates(waypoint.location()), waypoint.dimension().location())));
+            return worldWaypoints.size();
+        }
+
+        if (waypoints.isEmpty()) {
+            source.sendFeedback(Component.translatable("commands.cwaypoint.list.empty"));
+            return 0;
+        }
+
+        AtomicInteger count = new AtomicInteger();
```
Earthcomputer: I don't really like using `AtomicInteger` for the purpose of mutating inside a lambda. You can either use Apache's `MutableInteger` or what I normally do is use a length-1 array (`int[] count = {0}; ... count[0] += worldWaypoints.size()`). Alternatively, you can use a for loop to iterate over the map's `entrySet()` to avoid the issue.

src/main/java/net/earthcomputer/clientcommands/command/WaypointCommand.java
```diff
@@ -0,0 +1,413 @@
...
+        }
+    }
+
+    private static void loadFile() throws Exception {
+        waypoints.clear();
+        CompoundTag rootTag = NbtIo.read(ClientCommands.configDir.resolve("waypoints.dat"));
+        if (rootTag == null) {
+            return;
+        }
+        // TODO: update-sensitive: apply custom data fixes when it becomes necessary
```
Earthcomputer: I'm trying to remove these `// TODO: update-sensitive` comments nowadays and replace them with tests that can be run automatically. But I'll allow you to be lazy here and delete the comment without adding a test, I really think this file format is the least likely out of all the things we have which could silently break.

src/main/java/net/earthcomputer/clientcommands/command/WaypointCommand.java
```diff
@@ -0,0 +1,413 @@
...
+        Vector2d position = new Vector2d(cameraEntity.getEyePosition().x, cameraEntity.getEyePosition().z);
+
+        List<WaypointLabelLocation> xPositions = new ArrayList<>();
+        waypoints.forEach((waypointName, waypoint) -> {
+            if (!waypoint.dimension().location().equals(minecraft.level.dimension().location())) {
+                return;
+            }
+
+            double distanceSquared = waypoint.location().distToCenterSqr(cameraEntity.position());
+            long distance = Math.round(Math.sqrt(distanceSquared));
```
Earthcomputer, line 318: Is there a reason to call `distToCenterSqr` if you're immediately square rooting it? Is there not a `distanceToCenter` function?

xpple: There is not.

src/main/java/net/earthcomputer/clientcommands/ClientCommands.java
xpple: Sure.

src/main/java/net/earthcomputer/clientcommands/command/WaypointCommand.java
xpple: Why not keep the comment? Better than nothing right?

Earthcomputer (approved):

PR #514, opened 2023-03-28
Simplify code

xpple:
Funny, cause it actually reverts the changes from my PR from almost two years ago (#244).


Earthcomputer:
Does the `Box` constructor now do `min` and `max` internally?

Edit: yes it does

PR #517, opened 2023-04-02
Propose BetterConfig to support non temporary configs

### LINKED ISSUE #445: [Suggestion] Permanent settings changes?
It doesn't appear like there is a way to permanently change settings with a command. Maybe you could find a way to do this? Maybe /crule?

xpple:
Closes #445. Configs can now be accessed and changed through the `cconfig` command, replacing `ctemprule`. Let me know if this approach is okay.

Edit: The default config file would look like this, the rest of the configs remain temporary.
```json
{
  "maxChorusItemThrows": 2048,
  "toolBreakWarning": false,
  "maxEnchantItemThrows": 2048,
  "commandExecutionLimit": 65536,
  "acceptC2CPackets": false
}
```


erxson:
Thank you!

PR #524, opened 2023-04-25
Update zh_cn.json

GodGun968:


Earthcomputer:
The config translations are now in a separate mod: https://github.com/xpple/BetterConfig

Up to you if you want to do them as well :)

PR #527, opened 2023-05-15
Allow required custom resource packs to be bypassed

xpple:
As the title suggests, the player will be able to bypass the required custom resource pack often imposed by servers. For the newly added config `resourcePackBypassPolicy`, three values are distinguished:

1. `VANILLA` - the client will behave as usual,
2. `BYPASS_FORCED` - the client will automatically bypass _required_ custom resource packs,
3. `BYPASS_ALL` - the client will automatically bypass _all_ custom resource packs.

This config value will only apply when the player has not manually changed the resource pack policy in the server info screen. That is, if the player has set this policy to "disabled", the custom resource pack will always be bypassed. If the player has set this policy to "enabled", the custom resource pack will always be accepted.

If the policy is "prompt" (the default), the config is set to `VANILLA` and the custom resource pack is required, the confirmation screen will also contain a bypass button that when clicked will set the config to `BYPASS_FORCED` and bypass the required custom resource pack. In all other cases the default confirmation screen will be shown.

### DIFF
--- a/src/main/java/net/earthcomputer/clientcommands/features/BypassRequiredResourcePackScreen.java
+++ b/src/main/java/net/earthcomputer/clientcommands/features/BypassRequiredResourcePackScreen.java
@@ -1,0 +1,23 @@
+package net.earthcomputer.clientcommands.features;
+
+import it.unimi.dsi.fastutil.booleans.BooleanConsumer;
+import net.minecraft.client.gui.screen.ConfirmScreen;
+import net.minecraft.client.gui.widget.ButtonWidget;
+import net.minecraft.screen.ScreenTexts;
+import net.minecraft.text.Text;
+
+public class BypassRequiredResourcePackScreen extends ConfirmScreen {
+
+    private final Runnable bypassCallback;
+
+    public BypassRequiredResourcePackScreen(BooleanConsumer callback, Runnable bypassCallback, Text message) {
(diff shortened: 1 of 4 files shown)


src/main/java/net/earthcomputer/clientcommands/features/BypassRequiredResourcePackScreen.java
```diff
@@ -0,0 +1,23 @@
...
+
+    public BypassRequiredResourcePackScreen(BooleanConsumer callback, Runnable bypassCallback, Text message) {
+        super(callback, Text.translatable("multiplayer.requiredTexturePrompt.line1"), message, ScreenTexts.PROCEED, Text.translatable("menu.disconnect"));
+        this.bypassCallback = bypassCallback;
+    }
+
+    @Override
+    protected void addButtons(int y) {
+        super.addButtons(y);
+        this.addButton(ButtonWidget.builder(Text.translatable("resourcePackBypass.bypass"), button -> this.bypassCallback.run()).dimensions(this.width / 2 - 155 / 2, y + 30, 150, 20).build());
```
Starmania, line 21: They will be no problem on big resolution ?

Cyclopropinon:
why closed?

xpple:
I realised that this is definitely out of scope for clientcommands.

PR #567, opened 2023-09-15
Update zh_cn.json

GodGun968:


src/main/resources/assets/clientcommands/lang/zh_cn.json
```diff
@@ -107,6 +108,8 @@
   "commands.cglow.entity.success": "成功使%d个实体发光",
   "commands.cglow.area.success": "成功使%d个区域发光",
 
+  "commands.chat.success": "成功打开 c？？",
```
Earthcomputer: "c🎩" is a pun that only works in English. I'm not sure how you should translate that to Chinese to be honest.

GodGun968: What are the two meanings? Can you be more specific? I can try to translate it.

Earthcomputer: "chat" is like, the chat, i.e. the command opens the chat.
"c🎩" is "c" (the command prefix in clientcommands) + "hat" (the hat emoji).

GodGun968: Oh, It's not easy to translate
This command seems to be only available on April Fool's Day
Does it do anything?
Maybe I can add some funny sentences

Earthcomputer: Yeah, it just opens chat. Your updated translation looks good :)

PR #585, opened 2024-01-08
Implement `replace` functionality for `/cghostblock fill`

VelizarBG:
Yep.

### DIFF
--- a/src/main/java/net/earthcomputer/clientcommands/command/GhostBlockCommand.java
+++ b/src/main/java/net/earthcomputer/clientcommands/command/GhostBlockCommand.java
@@ -4,13 +4,18 @@ import com.mojang.brigadier.Command;
 import com.mojang.brigadier.CommandDispatcher;
 import com.mojang.brigadier.exceptions.CommandSyntaxException;
 import com.mojang.brigadier.exceptions.SimpleCommandExceptionType;
+import dev.xpple.clientarguments.arguments.CBlockPredicateArgumentType;
 import net.fabricmc.fabric.api.client.command.v2.FabricClientCommandSource;
 import net.minecraft.block.BlockState;
+import net.minecraft.block.pattern.CachedBlockPosition;
 import net.minecraft.client.world.ClientWorld;
 import net.minecraft.command.CommandRegistryAccess;
 import net.minecraft.text.Text;
 import net.minecraft.util.math.BlockBox;
 import net.minecraft.util.math.BlockPos;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.function.Predicate;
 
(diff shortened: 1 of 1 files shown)


src/main/java/net/earthcomputer/clientcommands/command/GhostBlockCommand.java
```diff
@@ -31,7 +36,10 @@ public static void register(CommandDispatcher<FabricClientCommandSource> dispatc
                 .then(argument("from", blockPos())
                     .then(argument("to", blockPos())
                         .then(argument("block", blockState(registryAccess))
-                            .executes(ctx -> fillGhostBlocks(ctx.getSource(), getCBlockPos(ctx, "from"), getCBlockPos(ctx, "to"), getCBlockState(ctx, "block").getBlockState())))))));
+                            .executes(ctx -> fillGhostBlocks(ctx.getSource(), getCBlockPos(ctx, "from"), getCBlockPos(ctx, "to"), getCBlockState(ctx, "block").getBlockState(), null))
+                            .then(literal("replace")
+                                .then(argument("filter", CBlockPredicateArgumentType.blockPredicate(registryAccess))
+                                    .executes(ctx -> fillGhostBlocks(ctx.getSource(), getCBlockPos(ctx, "from"), getCBlockPos(ctx, "to"), getCBlockState(ctx, "block").getBlockState(), CBlockPredicateArgumentType.getCBlockPredicate(ctx, "filter"))))))))));
```
Earthcomputer: You should use static imports for argument types

src/main/java/net/earthcomputer/clientcommands/command/GhostBlockCommand.java
```diff
@@ -49,7 +57,7 @@ private static int setGhostBlock(FabricClientCommandSource source, BlockPos pos,
         }
     }
 
-    private static int fillGhostBlocks(FabricClientCommandSource source, BlockPos from, BlockPos to, BlockState state) throws CommandSyntaxException {
+    private static int fillGhostBlocks(FabricClientCommandSource source, BlockPos from, BlockPos to, BlockState state, @Nullable Predicate<CachedBlockPosition> filter) throws CommandSyntaxException {
```
Earthcomputer: Any reason this predicate has to be nullable instead of using `pos -> true` instead of `null`?

Earthcomputer (changes requested):

src/main/java/net/earthcomputer/clientcommands/command/GhostBlockCommand.java
VelizarBG: Good point. It's just what I found in the vanilla replace code, but I can do that instead if you'd like.

Earthcomputer (approved):

PR #594, opened 2024-01-30
Add the cclearchat command

xpple:
The `clearHistory` argument only clears the history in memory, the history will be restored once the game is restarted.

### DIFF
--- a/src/main/java/net/earthcomputer/clientcommands/command/ClearChatCommand.java
+++ b/src/main/java/net/earthcomputer/clientcommands/command/ClearChatCommand.java
@@ -1,0 +1,26 @@
+package net.earthcomputer.clientcommands.command;
+
+import com.mojang.brigadier.Command;
+import com.mojang.brigadier.CommandDispatcher;
+import net.fabricmc.fabric.api.client.command.v2.FabricClientCommandSource;
+
+import static com.mojang.brigadier.arguments.BoolArgumentType.*;
+import static net.fabricmc.fabric.api.client.command.v2.ClientCommandManager.*;
+
+public class ClearChatCommand {
+    public static void register(CommandDispatcher<FabricClientCommandSource> dispatcher) {
+        dispatcher.register(literal("cclearchat")
+            .executes(ctx -> clearChat(ctx.getSource()))
+            .then(argument("clearHistory", bool())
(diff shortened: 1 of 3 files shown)


xpple:
Note to self: Change mixin so that the vanilla behaviour remains the same.

Earthcomputer:
What's the difference between this and F3+D?

xpple:
See [#dev-general > 💬](https://discord.com/channels/703363189081112647/703598752664912012/1202037976972918794).

src/main/java/net/earthcomputer/clientcommands/mixin/MixinChatHud.java
```diff
@@ -114,4 +117,9 @@ private static boolean handleC2CPacket(String content) {
         }
         return true;
     }
+
+    @Redirect(method = "clear", at = @At(value = "INVOKE", target = "Lnet/minecraft/util/collection/ArrayListDeque;addAll(Ljava/util/Collection;)Z"))
+    private boolean preventAddAll(ArrayListDeque<String> instance, Collection<String> collection) {
+        return false;
+    }
```
haykam821, line 124: Are you sure this mixin does not affect any other behavior?

src/main/java/net/earthcomputer/clientcommands/command/ClearChatCommand.java
```diff
@@ -0,0 +1,26 @@
...
+            .then(argument("clearHistory", bool())
+                .executes(ctx -> clearChat(ctx.getSource(), getBool(ctx, "clearHistory")))));
+    }
+
+    private static int clearChat(FabricClientCommandSource source) {
+        return clearChat(source, false);
+    }
+
+    private static int clearChat(FabricClientCommandSource source, boolean clearHistory) {
+        source.getClient().inGameHud.getChatHud().clear(clearHistory);
```
haykam821, line 23: You mentioned that <kbd>F3</kbd>+<kbd>3</kbd> and `/cclearchat false` both clear chat while leaving command history accessible while `/cclearchat true` clears chat and removes all command history for the current session. The former behavior is expected because chat isn't persisted across sessions, but should `/cclearchat true` also clear command history across sessions?

PR #595, opened 2024-02-02
Snake improvements

Gaming32:
Adds multiplayer snake. Recreating this PR from a different origin branch.


xpple:
Does this not face the same issue as before? That you will get kicked for sending too many packets unless you are `/op`'ed.

Gaming32:
I have no memory of what the issues were when I re-opened this. It's been over a year. But now I remember why it was Draft.

xpple:
@Gaming32 since unfortunately multiplayer snake is infeasible for the reason above, could you perhaps remove it and keep the improvements you have made?

PR #598, opened 2024-02-05
Add the clisten command

xpple:
Adds a command that lets the user listen to incoming or outgoing packets. The command supports two outputs that are configurable through the config `packetDumpMethod`:

1. `REFLECTION`. This output format displays the packet's data using reflection on the corresponding packet class. All fields of the packet will be recursively serialized to `Text` (`Component`) providing a nice display for the user.
2. `BYTE_BUF`. This output format displays the packet's data by keeping track of the calls to the methods in `PacketByteBuf` (`FriendlyByteBuf`) when the packet is serialized into it. The resulting output will be JSON. This packet dump method is written by @Gaming32.

An example of method 1:
image
An example of method 2:
image
Another example of method 1:
image
Another example of method 2:
image

Since protocol related analysis is more natural in mojmap ([wiki.vg/Protocol](https://wiki.vg/Protocol)), everything related to this command is in that mapping as well. This necessitated the `MappingsHelper` class to go back and forth between mappings. Perhaps in the future we can further use this class for a basic `/cmappings` command.


xpple:
I confirmed that both methods work in 1.20.5 (at least currently), see [xpple:clisten+1.20.5](https://github.com/xpple/clientcommands/tree/clisten%2B1.20.5).

xpple:
Confirmed to work after port to mojmap, no mappings issues in dev or prod.

src/main/java/net/earthcomputer/clientcommands/ReflectionUtils.java
```diff
@@ -0,0 +1,19 @@
+package net.earthcomputer.clientcommands;
+
+import java.lang.reflect.Field;
+import java.util.stream.Stream;
+
+public class ReflectionUtils {
```
Earthcomputer: This utility class could be final and have a private constructor

src/main/java/net/earthcomputer/clientcommands/UnsafeUtils.java
```diff
@@ -0,0 +1,34 @@
...
+
+    static {
+        try {
+            final Field unsafeField = Unsafe.class.getDeclaredField("theUnsafe");
+            unsafeField.setAccessible(true);
+            UNSAFE = (Unsafe) unsafeField.get(null);
+            final Field implLookupField = MethodHandles.Lookup.class.getDeclaredField("IMPL_LOOKUP");
+            IMPL_LOOKUP = (MethodHandles.Lookup) UNSAFE.getObject(UNSAFE.staticFieldBase(implLookupField), UNSAFE.staticFieldOffset(implLookupField));
+        } catch (Exception e) {
+            throw new RuntimeException(e);
```
Earthcomputer: This should probably more gracefully fail, and disable the command. It could possibly fail if the user is using a different JVM such as OpenJ9, or in future Java versions.

I'm not super keen on having this hack in here at all, but for now I've decided it's fine.

src/main/java/net/earthcomputer/clientcommands/command/ListenCommand.java
```diff
@@ -0,0 +1,269 @@
...
+
+import com.google.gson.stream.JsonWriter;
+import com.mojang.brigadier.Command;
+import com.mojang.brigadier.CommandDispatcher;
+import com.mojang.brigadier.Message;
+import com.mojang.brigadier.exceptions.CommandSyntaxException;
+import com.mojang.brigadier.exceptions.SimpleCommandExceptionType;
+import it.unimi.dsi.fastutil.objects.ReferenceOpenHashSet;
+import it.unimi.dsi.fastutil.objects.ReferenceSet;
+import net.earthcomputer.clientcommands.*;
```
Earthcomputer: Wildcard import

src/main/java/net/earthcomputer/clientcommands/command/ListenCommand.java
```diff
@@ -0,0 +1,269 @@
...
+import net.minecraft.world.level.ChunkPos;
+
+import java.io.IOException;
+import java.io.StringWriter;
+import java.lang.invoke.VarHandle;
+import java.lang.reflect.Array;
+import java.lang.reflect.InaccessibleObjectException;
+import java.lang.reflect.Modifier;
+import java.time.Instant;
+import java.util.*;
```
Earthcomputer: Wildcard import again, you should change your wildcard import limit in intellij to something large like 99999

src/main/java/net/earthcomputer/clientcommands/command/ListenCommand.java
```diff
@@ -0,0 +1,269 @@
...
+public class ListenCommand {
+
+    private static final SimpleCommandExceptionType ALREADY_LISTENING_EXCEPTION = new SimpleCommandExceptionType(Component.translatable("commands.clisten.add.failed"));
+    private static final SimpleCommandExceptionType NOT_LISTENING_EXCEPTION = new SimpleCommandExceptionType(Component.translatable("commands.clisten.remove.failed"));
+
+    private static final Set<Class<Packet<?>>> packets = new HashSet<>();
+
+    private static PacketCallback callback;
+
+    private static final ThreadLocal<ReferenceSet<Object>> SEEN = ThreadLocal.withInitial(ReferenceOpenHashSet::new);
```
Earthcomputer: It's not necessarily bad, but why the use of fastutil here over a regular `HashSet`?

src/main/java/net/earthcomputer/clientcommands/command/ListenCommand.java
```diff
@@ -0,0 +1,269 @@
...
+    private static int list(FabricClientCommandSource source) {
+        int amount = packets.size();
+        if (amount == 0) {
+            source.sendFeedback(Component.translatable("commands.clisten.list.none"));
+        } else {
+            source.sendFeedback(Component.translatable("commands.clisten.list"));
+            packets.forEach(packetClass -> {
+                String packetClassName = packetClass.getName().replace('.', '/');
+                Optional<String> mojmapName = MappingsHelper.namedOrIntermediaryToMojmap_class(packetClassName);
+                source.sendFeedback(Component.literal(mojmapName.orElseThrow().substring(MOJMAP_PACKET_PREFIX.length())));
```
Earthcomputer: Maybe just get the name after the package, in case Mojang decides to repackage their packets

src/main/java/net/earthcomputer/clientcommands/command/ListenCommand.java
```diff
@@ -0,0 +1,269 @@
...
+            return Component.empty();
+        } finally {
+            SEEN.get().remove(object);
+            if (SEEN.get().isEmpty()) {
+                SEEN.remove();
+            }
+        }
+    }
+
+    private static Component serializeInner(Object object) {
```
Earthcomputer: I am worried that Mojang might start using a different JDK class that we might want to pretty print (like you already had to add `Date` and `Instant`). This was the maintenance burden I was referring to in Discord. But at least we gracefully fallback to listing the fields in this case.

src/main/java/net/earthcomputer/clientcommands/command/ListenCommand.java
```diff
@@ -0,0 +1,269 @@
...
+    private static int clear(FabricClientCommandSource source) {
+        int amount = packets.size();
+        packets.clear();
+        source.sendFeedback(Component.translatable("commands.clisten.clear"));
+        return amount;
+    }
+
+    private static Component serialize(Object object) {
+        try {
+            if (SEEN.get().add(object)) {
```
Earthcomputer: Instead of using global state (even if it is `ThreadLocal`), would it be simpler and more efficient to pass around the `seen` set as a parameter in the recursive function? Also, perhaps there could be a configurable maximum depth after which it stops listing too deeply nested fields.

src/main/java/net/earthcomputer/clientcommands/command/arguments/MojmapPacketClassArgumentType.java
```diff
@@ -0,0 +1,68 @@
...
+import java.util.concurrent.CompletableFuture;
+import java.util.stream.Collectors;
+
+public class MojmapPacketClassArgumentType implements ArgumentType<Class<Packet<?>>> {
+
+    private static final Collection<String> EXAMPLES = Arrays.asList("ClientboundPlayerChatPacket", "ClientboundSystemChatMessage", "ServerboundContainerSlotStateChangedPacket");
+
+    public static final String MOJMAP_PACKET_PREFIX = "net/minecraft/network/protocol/game/";
+
+    private static final Set<String> mojmapPackets = MappingsHelper.mojmapClasses().stream()
```
Earthcomputer: Instead of doing it this way, you can get a list of packet classes that are in use via the `ConnectionProtocol` enum. This way you are also not depending on Mojang's packet package prefix, which is good if they decide to change it. This may require one or two access widener entries, but I still prefer that over doing it this way.

src/main/java/net/earthcomputer/clientcommands/features/MappingsHelper.java
```diff
@@ -0,0 +1,226 @@
...
+
+    private static final MemoryMappingTree officialIntermediaryNamed = new MemoryMappingTree();
+    private static final int SRC_INTERMEDIARY = 0;
+    private static final int DEST_INTERMEDIARY = 0;
+    private static final int SRC_NAMED = 1;
+    private static final int DEST_NAMED = 1;
+
+    private static final boolean IS_DEV_ENV = FabricLoader.getInstance().isDevelopmentEnvironment();
+
+    private static final HttpClient httpClient = HttpClient.newHttpClient();
```
Earthcomputer: This could be a local variable, just so the resources for the http client are cleaned up after they are no longer needed.

src/main/java/net/earthcomputer/clientcommands/features/MappingsHelper.java
```diff
@@ -0,0 +1,226 @@
...
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    public static Collection<? extends MappingTree.ClassMapping> mojmapClasses() {
+        return mojmapOfficial.getClasses();
+    }
+
+    public static Optional<String> mojmapToOfficial_class(String mojmapClass) {
```
Earthcomputer: Just fyi, I prefer using `@Nullable String` (with the annotation) over `Optional<String>`, as it makes for less verbose code and the annotation makes the IDE warn if you forget to check for null anyway. You can add this to the contribution guidelines if you feel like it.

src/main/java/net/earthcomputer/clientcommands/features/MappingsHelper.java
```diff
@@ -0,0 +1,226 @@
...
+                    });
+            });
+
+        try (InputStream stream = FabricLoader.class.getClassLoader().getResourceAsStream("mappings/mappings.tiny")) {
+            if (stream == null) {
+                throw new IOException("Could not find mappings.tiny");
+            }
+            MappingReader.read(new InputStreamReader(stream), IS_DEV_ENV ? MappingFormat.TINY_2_FILE : MappingFormat.TINY_FILE, officialIntermediaryNamed);
+        } catch (IOException e) {
+            throw new RuntimeException(e);
```
Earthcomputer: Exceptions should be handled gracefully here. What if Fabric loader decides to change where they store the mappings? We don't want to crash the whole game if this one command can't be used.

src/main/java/net/earthcomputer/clientcommands/features/MappingsHelper.java
```diff
@@ -0,0 +1,226 @@
...
+                            .uri(URI.create(mappingsUrl))
+                            .GET()
+                            .timeout(Duration.ofSeconds(5))
+                            .build();
+                        httpClient.sendAsync(mappingsRequest, HttpResponse.BodyHandlers.ofString())
+                            .thenApply(HttpResponse::body)
+                            .thenApply(StringReader::new)
+                            .thenAccept(reader -> {
+                                try {
+                                    MappingReader.read(reader, MappingFormat.PROGUARD_FILE, mojmapOfficial);
```
Earthcomputer: Is `MemoryMappingTree` thread safe? It's quite unlikely that the player manages to log into a world before this finishes, but if they do and run the command this will cause a race condition. Also in the future we might decide to use this class elsewhere that gets called earlier. The various access methods below should block while this request is still finishing (or until the request fails, if the case may be).

The mappings should probably be downloaded to a local file which could be used as a fallback in case the user later runs the game without an internet connection. Also this needs to fail gracefully if the user runs the game offline.

src/main/java/net/earthcomputer/clientcommands/features/PacketDumper.java
```diff
@@ -0,0 +1,701 @@
...
+import com.mojang.serialization.JsonOps;
+import com.mojang.util.ByteBufferTypeAdapter;
+import com.mojang.util.InstantTypeAdapter;
+import com.mojang.util.UUIDTypeAdapter;
+import io.netty.buffer.ByteBuf;
+import io.netty.buffer.Unpooled;
+import io.netty.handler.codec.EncoderException;
+import it.unimi.dsi.fastutil.ints.IntList;
+import net.minecraft.Util;
+import net.minecraft.core.*;
```
Earthcomputer: Wildcard import

src/main/java/net/earthcomputer/clientcommands/features/PacketDumper.java
```diff
@@ -0,0 +1,701 @@
...
+import java.io.UncheckedIOException;
+import java.nio.ByteBuffer;
+import java.nio.channels.FileChannel;
+import java.nio.channels.ScatteringByteChannel;
+import java.nio.charset.Charset;
+import java.security.PublicKey;
+import java.text.DateFormat;
+import java.text.SimpleDateFormat;
+import java.time.Instant;
+import java.util.*;
```
Earthcomputer: Wildcard import

src/main/java/net/earthcomputer/clientcommands/features/PacketDumper.java
```diff
@@ -0,0 +1,701 @@
...
+        writer.beginArray();
+        try {
+            packet.write(new PacketDumpByteBuf(writer));
+        } catch (UncheckedIOException e) {
+            throw e.getCause();
+        }
+        writer.endArray();
+    }
+
+    public static class PacketDumpByteBuf extends FriendlyByteBuf {
```
Earthcomputer: Does this class need to be public?

build.gradle
```diff
@@ -60,6 +60,8 @@ dependencies {
 	modRuntimeOnly('me.djtheredstoner:DevAuth-fabric:1.1.0') {
 		exclude group: 'net.fabricmc', module: 'fabric-loader'
 	}
+
+	include api('net.fabricmc:mapping-io:0.5.1')
```
Earthcomputer, line 64: I'm worried that this may clash with a version of mapping-io that is used inside Fabric loader. Does Fabric loader also JiJ it or does it just include it in the jar? If it's JiJ it should be fine as loader will pick the latest version, otherwise I'm worried as we don't want to break loader if it requires a later version of mapping-io. I'm also worried because mapping-io is still in a 0.* phase, which means 0.6 could contain breaking changes, which could break our mod on future Fabric loader versions, which is a problem because new loader versions can be used on old mc versions. To solve this, we could repackage mapping-io.

Earthcomputer (changes requested):

src/main/java/net/earthcomputer/clientcommands/command/ListenCommand.java
Gaming32: Because that will perform the hash of the object, which could be really slow. `ReferenceSet`s use the identity hash.

build.gradle
Gaming32: Loader shadows, relocates, and minimizes MIO.

src/main/java/net/earthcomputer/clientcommands/command/ListenCommand.java
xpple: I added `Date` only because `Component#translationArg(Date)` exists and I added `Instant` because the result looks nicer as a `Date`. I don't think JDK classes are a problem anymore with this `IMPL_LOOKUP` solution.

src/main/java/net/earthcomputer/clientcommands/features/MappingsHelper.java
xpple: I initially had that actually. But `Optional`s make the code a bit nicer because I use `Optional.ofNullable` quite a lot.

src/main/java/net/earthcomputer/clientcommands/UnsafeUtils.java
```diff
@@ -0,0 +1,54 @@
...
+ * @author Gaming32
+ */
+public final class UnsafeUtils {
+
+    private UnsafeUtils() {
+    }
+
+    private static final Logger LOGGER = LogUtils.getLogger();
+
+    private static final Unsafe UNSAFE = Util.make(() -> {
```
Earthcomputer: This field should be `@Nullable`

src/main/java/net/earthcomputer/clientcommands/UnsafeUtils.java
```diff
@@ -0,0 +1,54 @@
...
+            final Field unsafeField = Unsafe.class.getDeclaredField("theUnsafe");
+            unsafeField.setAccessible(true);
+            return (Unsafe) unsafeField.get(null);
+        } catch (Exception e) {
+            LOGGER.error("Could not access theUnsafe", e);
+            return null;
+        }
+    });
+
+    private static final MethodHandles.Lookup IMPL_LOOKUP = Util.make(() -> {
```
Earthcomputer: `@Nullable`

src/main/java/net/earthcomputer/clientcommands/UnsafeUtils.java
```diff
@@ -0,0 +1,54 @@
...
+            }
+            final Field implLookupField = MethodHandles.Lookup.class.getDeclaredField("IMPL_LOOKUP");
+            return (MethodHandles.Lookup) UNSAFE.getObject(UNSAFE.staticFieldBase(implLookupField), UNSAFE.staticFieldOffset(implLookupField));
+        } catch (Exception e) {
+            LOGGER.error("Could not access IMPL_LOOKUP", e);
+            return null;
+        }
+    });
+
+    public static Unsafe getUnsafe() {
```
Earthcomputer: `@Nullable`

src/main/java/net/earthcomputer/clientcommands/UnsafeUtils.java
```diff
@@ -0,0 +1,54 @@
...
+        return UNSAFE;
+    }
+
+    public static MethodHandles.Lookup getImplLookup() {
```
Earthcomputer: `@Nullable`

src/main/java/net/earthcomputer/clientcommands/command/ListenCommand.java
```diff
@@ -0,0 +1,285 @@
...
+import java.util.Optional;
+import java.util.Set;
+import java.util.UUID;
+
+import static net.earthcomputer.clientcommands.command.arguments.MojmapPacketClassArgumentType.*;
+import static net.fabricmc.fabric.api.client.command.v2.ClientCommandManager.*;
+
+public class ListenCommand {
+
+    public static boolean isEnabled = true;
```
Earthcomputer:
```suggestion
    public static volatile boolean isEnabled = true;
```

src/main/java/net/earthcomputer/clientcommands/ClientCommands.java
```diff
@@ -161,6 +165,9 @@ public static void registerCommands(CommandDispatcher<FabricClientCommandSource>
         WeatherCommand.register(dispatcher);
         PluginsCommand.register(dispatcher);
         CGameModeCommand.register(dispatcher);
+        if (ListenCommand.isEnabled) {
```
Earthcomputer: I've just thought of another race condition. This command could be registered and then the `isEnabled` boolean set to false afterwards. On Quilt, client-side commands are registered on startup as well for some reason that isn't clear to me. Might be a better idea to always register the command and check the boolean when it's executed.

Earthcomputer (approved):

PR #601, opened 2024-02-12
Add the cgamemode command

xpple:


Earthcomputer:
Perhaps `/cgamemode player` could be renamed to `/cgamemode get` or `/cgamemode query` to make it clear that you are not setting the gamemode?

Earthcomputer:
Although for advanced clientcommands users, I could be wrong, but the gamemode was already possible to query using `/cgetdata`

PR #609, opened 2024-03-15
Fixed support for ViaFabricPlus 3.1.0 and newer versions

florianreuth:
Added support for the new VFP version / the new API changes from VV, once 1.20.5 releases, the old code can be deleted.

### DIFF
--- a/src/main/java/net/earthcomputer/clientcommands/MultiVersionCompat.java
+++ b/src/main/java/net/earthcomputer/clientcommands/MultiVersionCompat.java
@@ -5,6 +5,9 @@ import com.mojang.logging.LogUtils;
 import io.netty.channel.Channel;
 import io.netty.channel.ChannelHandler;
 import net.fabricmc.loader.api.FabricLoader;
+import net.fabricmc.loader.api.ModContainer;
+import net.fabricmc.loader.api.Version;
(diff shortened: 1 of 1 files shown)


src/main/java/net/earthcomputer/clientcommands/MultiVersionCompat.java
```diff
@@ -40,10 +43,11 @@ public boolean doesItemExist(Item item) {
     public static final MultiVersionCompat INSTANCE = Util.make(() -> {
         try {
             FabricLoader loader = FabricLoader.getInstance();
-            if (loader.isModLoaded("viafabric")) {
+            ModContainer modContainer;
+            if (loader.getModContainer("viafabric").orElse(null) != null) {
```
crosby-moe: why use `loader.getModContainer("viafabric").orElse(null) != null` if you dont capture the `ModContainer`?

florianreuth: oh yeah right, I can probably just use isModLoaded() here

florianreuth: Fixed

florianreuth:
Applied Earth's suggestion from dc and splitted VFP support into two class files

PR #615, opened 2024-03-29
Fix crash when typing "/ " in chat

### LINKED ISSUE #610: Crash via indexing out of array
Typing "/\<space\>" leads to a crash in MixinChatScreen.java's onEdited function as value.substring(1).split(" ") returns a null string

DeltAndy123:
Fixes https://github.com/Earthcomputer/clientcommands/issues/610


Earthcomputer (changes requested):
Probably a cleaner fix would be to assign the array to a variable and check its length before getting the first string.

Earthcomputer:
Merged with changes, thanks!

PR #632, opened 2024-04-30
Add multiplayer tic-tac-toe

### LINKED ISSUE #629: suggestion: /ctictactoe command
playing tictactoe through c2c packets

xpple:
Closes #629.


src/main/java/net/earthcomputer/clientcommands/command/ClientCommandHelper.java
```diff
@@ -77,8 +81,19 @@ public static Component getCommandTextComponent(@Translatable String translation
...
-                .withHoverEvent(new HoverEvent(HoverEvent.Action.SHOW_TEXT, Component.literal(command))));
+            .withClickEvent(new ClickEvent(ClickEvent.Action.RUN_COMMAND, command))
+            .withHoverEvent(new HoverEvent(HoverEvent.Action.SHOW_TEXT, Component.literal(command))));
     }
 
+    public static final Map<String, Runnable> runnables = new HashMap<>();
+
+    public static String registerCode(Runnable code) {
+        String randomString = new Random().ints(48, 122 + 1)
+            .filter(i -> (i <= 57 || i >= 65) && (i <= 90 || i >= 97))
```
Earthcomputer: These can be char literals rather than magic constants. Although a better idea is to pick a random number between 0 and 26 + 26 + 10 and convert that to a lower- or uppercase letter or number with an if-else chain.

src/main/java/net/earthcomputer/clientcommands/command/PingCommand.java
```diff
@@ -49,23 +48,16 @@ private static int printPing(FabricClientCommandSource source, Collection<GamePr
             throw PLAYER_IN_SINGLEPLAYER_EXCEPTION.create();
         }
 
-        if (profiles.isEmpty()) {
+        if (profiles.size() != 1) {
             throw PLAYER_NOT_FOUND_EXCEPTION.create();
         }
-        if (profiles.size() > 1) {
-            throw TOO_MANY_PLAYERS_EXCEPTION.create();
-        }
```
Earthcomputer, line 57: I think it's better to keep these as separate exceptions for UX reasons. If you target multiple players it doesn't make sense to say "no player found".

src/main/resources/assets/clientcommands/lang/en_us.json
```diff
@@ -340,5 +347,14 @@
...
+  "c2cpacket.messageC2CPacket.outgoing": "you -> %s: %s",
+
+  "c2cpacket.startTicTacToeGameC2CPacket.outgoing.invited": "You invited %s to a game of tic tac toe",
+  "c2cpacket.startTicTacToeGameC2CPacket.incoming": "%s invited you to a game of tic tac toe",
+  "c2cpacket.startTicTacToeGameC2CPacket.incoming.accept": "Accept",
+  "c2cpacket.startTicTacToeGameC2CPacket.incoming.accept.hover": "Click to accept",
+  "c2cpacket.startTicTacToeGameC2CPacket.incoming.accepted": "%s has accepted your invitation, click to make your move",
+  "c2cpacket.startTicTacToeGameC2CPacket.outgoing.accept": "Accepted the invitation, your opponent will go first",
+  "c2cpacket.putTicTacToeMarkC2CPacket.incoming.lost": "%s has made a move in tic tac toe and won",
+  "c2cpacket.putTicTacToeMarkC2CPacket.incoming": "%s has made a move in tic tac toe, click to make your move"
```
Earthcomputer: tic-tac-toe is hyphenated

Earthcomputer (changes requested):

src/main/java/net/earthcomputer/clientcommands/command/PingCommand.java
xpple: It's nonsensical to query more than one game profile, I'd even say that the argument type returning a `Collection` is a quirk.

Earthcomputer: Does it not return multiple values if you use a target selector?

xpple: It does, but for the purposes of this command (and every other command using this argument type), it does not make sense to want that. If the query does result in more than one profile, it makes sense to say that _the_ player was not found, based on the target selector given by the user who expected a single result. I suppose this is also the reason Mojang does not use this argument type internally anymore at all, and now uses `EntityArgument.player()`. We could of course reflect this change, but we chose not to (see [this conversation](https://discord.com/channels/703363189081112647/703598704266969099/1232687209136656415)).

Earthcomputer: Just because it's nonsensical to do doesn't mean there shouldn't be a proper error message for it.

xpple: So you're saying we should add a `TOO_MANY_PLAYERS_EXCEPTION` everywhere where we use this argument type?

Earthcomputer: Yes. If that becomes a problem we can add a helper method but afaik this is the only place we use it.

xpple: We use it in lots of places, and importantly without this extra exception.

Earthcomputer (changes requested):
Please re-add the removed exception, and if we've forgotten to use it in other places add it there too (maybe with a helper method).

src/main/java/net/earthcomputer/clientcommands/command/PingCommand.java
xpple: I have replaced every usage of `CGameProfileArgument.gameProfile()` with `CEntityArgument.player()`. This allows for the same functionality but does not allow multiple matches when an entity selector is used.

xpple: It actually turns out the `profiles.isEmpty()` check we did everywhere was unnecessary, that is already prevented by the argument type 😄.

Earthcomputer (approved):
@xpple you can merge this if it's ready

src/main/java/net/earthcomputer/clientcommands/command/TicTacToeCommand.java
```diff
@@ -0,0 +1,280 @@
...
+                    this.board[x][y] = mark;
+                    this.yourTurn = !this.yourTurn;
+                    return true;
+                }
+            }
+            return false;
+        }
+
+        public Mark getWinner() {
+            for (byte x = 0; x < 3; x++) {
```
RealRTTV, line 164: Why is byte used here instead of int?

src/main/java/net/earthcomputer/clientcommands/command/TicTacToeCommand.java
```diff
@@ -0,0 +1,280 @@
...
+        private static final ResourceLocation MARKS_TEXTURE = new ResourceLocation("clientcommands", "textures/tic_tac_toe/marks.png");
+
+        private static final int gridSizeTexture = 512;
+        private static final int markSizeTexture = 152;
+
+        private static final int gridSize = 256;
+        private static final int cellSize = 80;
+        private static final int borderSize = 8;
+        private static final int markSize = 76;
+        private static final int padding = 2;
```
RealRTTV: Can these static values be in uppercase camel case?

src/main/java/net/earthcomputer/clientcommands/command/TicTacToeCommand.java
```diff
@@ -0,0 +1,280 @@
...
+                    guiGraphics.blit(MARKS_TEXTURE, startX + (cellSize + borderSize) * x + padding, startY + (cellSize + borderSize) * y + padding, markSize, markSize, offset, 0, markSizeTexture, markSizeTexture, 2 * markSizeTexture, markSizeTexture);
+                }
+            }
+        }
+
+        @Override
+        public boolean mouseClicked(double mouseX, double mouseY, int button) {
+            int startX = (this.width - gridSize) / 2;
+            int startY = (this.height - gridSize) / 2;
+            if (mouseX < startX || mouseX > startX + gridSize || mouseY < startY || mouseY > startY + gridSize) {
```
RealRTTV: Personal preference but I'd rather use `if (!((mouseX >= startX && mouseX <= startX + gridSize && mouseY >= startY && mouseY <= startY + gridSize)))` as it's (personally) easier to understand it's a guard clause form of checking if you're within the bounds of the board.

src/main/java/net/earthcomputer/clientcommands/command/TicTacToeCommand.java
xpple: Idk because it could be 😄. It's also a byte in the packet.

src/main/java/net/earthcomputer/clientcommands/command/TicTacToeCommand.java
```diff
@@ -0,0 +1,280 @@
...
+                for (byte y = 0; y < 3; y++) {
+                    TicTacToeGame.Mark mark = board[x][y];
+                    if (mark == null) {
+                        continue;
+                    }
+                    int offset = switch (mark) {
+                        case NOUGHT -> 0;
+                        case CROSS -> markSizeTexture;
+                    };
+                    guiGraphics.blit(MARKS_TEXTURE, startX + (cellSize + borderSize) * x + padding, startY + (cellSize + borderSize) * y + padding, markSize, markSize, offset, 0, markSizeTexture, markSizeTexture, 2 * markSizeTexture, markSizeTexture);
```
RealRTTV: I'd consider refactoring the uv coordinates and texture coordinates to use constants here

src/main/java/net/earthcomputer/clientcommands/command/TicTacToeCommand.java
xpple: No, but they can be SCREAMING_SNAKE_CASE ;).

src/main/java/net/earthcomputer/clientcommands/command/TicTacToeCommand.java
xpple: I disagree. Currently it guards against the case when the mouse is outside of the grid, which I think is easy to understand.

src/main/java/net/earthcomputer/clientcommands/command/TicTacToeCommand.java
xpple: Well I am already using constants.

RealRTTV: i mean adding
`static final MARKS_TEXTURE_WIDTH`
`static final MARKS_TEXTURE_HEIGHT`
so that you won't propagate issues if you change the texture dimensions to add more textures to `marks.png`

RealRTTV: sorry, was unclear

Earthcomputer (approved):

src/main/java/net/earthcomputer/clientcommands/command/TicTacToeCommand.java
xpple: I think it's fine, there will only ever be two textures (noughts and crosses, no more), which is a small enough number to be hardcoded.

xpple:
Not merging yet, we will probably have to think about users having different versions of clientcommands. This is because they might have different C2C packets, which will result in one of the users disconnecting upon receiving an unknown packet. This can be abused, possibly forcing someone to disable C2C packets.

Earthcomputer (approved):

PR #635, opened 2024-05-24
Add --keep-searching option to /cfindblock

### LINKED ISSUE #182: Add --keep-searching option to /cfindblock
For example, `/cfindblock spawner --i 20` would run the command every 20 seconds. Not sure how many commands this could be realistically applied to

RealRTTV:
Closes #182


RealRTTV:
im gonna change the chunks to a LinkedHashSet<ChunkPos> so that i can pop them in the most efficient order

xpple:
I very much like the progress bar you implemented, but could it be a separate PR?

RealRTTV:
yeah sure, i forgot how git works, ill do it later today

RealRTTV:
done, lgtm

src/main/java/net/earthcomputer/clientcommands/command/FindBlockCommand.java
```diff
@@ -73,6 +78,13 @@ protected boolean canScanChunkSection(Entity cameraEntity, SectionPos pos) {
             return hasBlockState(pos, predicate::canEverMatch) && super.canScanChunkSection(cameraEntity, pos);
         }
 
+        @Override
+        public boolean condition() {
+            boolean condition = hasChunksRemaining() || (keepSearching && closestBlock == null);
```
Earthcomputer: The reason for this override is to stop searching when a block has been found and there are no more loaded chunks to search. I think this code could be cleaner if you set `keepSearching` to `false` after you find the block. This has the same effect and is logically marginally clearer IMO. It also removes the need for this override.

src/main/java/net/earthcomputer/clientcommands/command/SignSearchCommand.java
```diff
@@ -30,10 +29,10 @@ public static void register(CommandDispatcher<FabricClientCommandSource> dispatc
         dispatcher.register(literal("csignsearch")
             .then(literal("text")
                 .then(argument("query", greedyString())
-                    .executes(ctx -> FindBlockCommand.findBlock(Component.translatable("commands.csignsearch.starting"), predicate(getString(ctx, "query"))))))
+                    .executes(ctx -> FindBlockCommand.findBlock(ctx, Component.translatable("commands.csignsearch.starting"), predicate(getString(ctx, "query"))))))
             .then(literal("regex")
                 .then(argument("query", greedyRegex())
-                    .executes(ctx -> FindBlockCommand.findBlock(Component.translatable("commands.csignsearch.starting"), predicate(getRegex(ctx, "query")))))));
+                    .executes(ctx -> FindBlockCommand.findBlock(ctx, Component.translatable("commands.csignsearch.starting"), predicate(getRegex(ctx, "query")))))));
```
Earthcomputer, line 34: We could maybe add a `--keep-searching` option to `/csignsearch` as well while we're at it.

src/main/java/net/earthcomputer/clientcommands/task/ItemThrowTask.java
```diff
@@ -66,7 +66,6 @@ protected void onTick() {
                 onFailedToThrowItem();
                 return;
             }
-            sentItemThrows++;
```
Earthcomputer: ?

src/main/java/net/earthcomputer/clientcommands/task/RenderDistanceScanTask.java
```diff
@@ -17,28 +18,37 @@
...
+        remainingChunks = new LinkedHashSet<>();
+        Entity cameraEntity = Minecraft.getInstance().cameraEntity;
+        if (cameraEntity == null) {
+            _break();
+            return;
+        }
+        BlockPos.spiralAround(new BlockPos(Mth.floor(cameraEntity.getX()) >> 4, 0, Mth.floor(cameraEntity.getZ()) >> 4), Minecraft.getInstance().options.renderDistance().get(), Direction.EAST, Direction.SOUTH).iterator().forEachRemaining(pos -> remainingChunks.add(new ChunkPos(pos.getX(), pos.getZ())));
+        ClientLevelEvents.CHUNK_UPDATE.register((level, pos, oldState, newState) -> remainingChunks.add(new ChunkPos(pos)));
+        ClientLevelEvents.UNLOAD_CHUNK.register((level, pos) -> remainingChunks.remove(pos));
+        ClientLevelEvents.LOAD_CHUNK.register((level, pos) -> remainingChunks.add(pos));
```
Earthcomputer: A few of points here:
- You are registering these listeners without ever unregistering them, leading to a memory leak. Take a look at `ItemThrowTask` for a way to get tasks to listen to events.
- As with `ItemThrowTask`, you probably don't want multiple render distance scan tasks from happening at the same time, so you could add it to the mutex keys (see `ItemThrowTask`).
- These events unconditionally add to the remaining chunks, whereas you probably only want to do that if `keepSearching` is enabled.
- If a block is changed, you may as well check if the changed block matches the block you're looking for (via `scanBlock`), rather than adding the entire chunk to be re-scanned. It does affect the order they're scanned in but I'm not so fussed about that, the main reason it's scanned in this order is to return the found block sooner if it's closer to the player.

Earthcomputer (changes requested):

src/main/java/net/earthcomputer/clientcommands/command/SignSearchCommand.java
RealRTTV: oh yeah, never even thought about that

src/main/java/net/earthcomputer/clientcommands/task/ItemThrowTask.java
RealRTTV: yeah idk why this was removed, sorry

src/main/java/net/earthcomputer/clientcommands/task/RenderDistanceScanTask.java
RealRTTV: - I was aware of the memory leak, thought I fixed it
- didn't think of that, thanks
- bruh
- due to the `throws CommandSyntaxException` of `doTick()`, I'd need to handle that in my lambda, what should I do?

Earthcomputer: - I'm not sure what you mean by "bruh". Without `keepSearching` you only want to search once.
- You should try-catch it and report the error to the user, the same as what `onTick()` does.

RealRTTV: - i didn't notice that i didn't add that
- ok sounds good

RealRTTV: I can't do what ItemThrowTask does because you can run multiple search tasks at once.

Earthcomputer: That's what the mutex key was for.

src/main/java/net/earthcomputer/clientcommands/task/RenderDistanceScanTask.java
```diff
@@ -16,29 +18,62 @@
...
-        squarePosIterator = createIterator();
+        remainingChunks = new LongLinkedOpenHashSet();
+        Entity cameraEntity = Minecraft.getInstance().cameraEntity;
+        if (cameraEntity == null) {
+            _break();
+            return;
+        }
+        BlockPos.spiralAround(new BlockPos(Mth.floor(cameraEntity.getX()) >> 4, 0, Mth.floor(cameraEntity.getZ()) >> 4), Minecraft.getInstance().options.renderDistance().get(), Direction.EAST, Direction.SOUTH).iterator().forEachRemaining(pos -> remainingChunks.add(ChunkPos.asLong(pos.getX(), pos.getZ())));
+        ClientLevelEvents.UNLOAD_CHUNK.register((level, pos) -> remainingChunks.remove(pos.toLong()));
+        ClientLevelEvents.LOAD_CHUNK.register((level, pos) -> remainingChunks.add(pos.toLong()));
```
Earthcomputer: These events should be registered in the static initializer too. The weak reference should just be a reference to the task, keep it simple.

Earthcomputer (changes requested):

src/main/java/net/earthcomputer/clientcommands/task/RenderDistanceScanTask.java
```diff
@@ -16,29 +18,70 @@
...
+import java.util.Set;
 import java.util.function.Predicate;
 
 public abstract class RenderDistanceScanTask extends SimpleTask {
     private static final long MAX_SCAN_TIME = 30_000_000L; // 30ms
+    private static final Set<Object> MUTEX_KEYS = Set.of(RenderDistanceScanTask.class);
+
+    static {
+        ClientLevelEvents.CHUNK_UPDATE.register((level, pos, oldState, newState) -> {
+            @Nullable WeakReference<RenderDistanceScanTask> currentScanTask = RenderDistanceScanTask.currentScanTask;
```
Earthcomputer: Nit: there is no need to annotate local variables with `@Nullable`, IntelliJ is smart enough to infer it in most cases. Otherwise looks good.

Earthcomputer (changes requested):

src/main/java/net/earthcomputer/clientcommands/task/RenderDistanceScanTask.java
RealRTTV: i can't access my computer rn, if you have a commit changes I’ll add it rq

Earthcomputer (approved):

PR #636, opened 2024-05-25
Add elytra-swapping

RealRTTV:
If you have an elytra in your inventory and press space mid-air, it'll swap to that elytra, and if you touch the ground with an elytra, it'll grab the first armor item in your inventory. Additionally, if you press space mid-air, your elytra will swap with that armor piece.


xpple:
I am tempted to close this because it is not a command and besides probably out of scope for clientcommands.

RealRTTV:
infiniteTools isn't a command

xpple:
Indeed, but that is arguably more in scope. clientcommands isn't a general utility mod.

PR #649, opened 2024-06-23
Fix recursive alias detection

### LINKED ISSUE #645: Cant run alias command after running it once
You cant run a alias after running it once

## steps to reproduce
1. make a alias via the `calias` command
2. run the alias twice

mygizli04:
Fixes #645 

When trying to detect whether or not an alias is recursive, it's added to an array of seen aliases. However this array was never cleared meaning after an alias is executed, it cannot be executed again with the error 'Recursive alias "[name of alias]"' appearing in chat despite there being no recursion.

I've confirmed that this issue does not occur on 2.8.18 but occurs after 2.8.19, after the recursion detection was added.

This pull request fixes the issue by removing the alias from the list of seen aliases after the command is completed.

### DIFF
--- a/src/main/java/net/earthcomputer/clientcommands/command/AliasCommand.java
+++ b/src/main/java/net/earthcomputer/clientcommands/command/AliasCommand.java
@@ -78,40 +78,45 @@ public class AliasCommand {
         if (cmd == null) {
             throw NOT_FOUND_EXCEPTION.create(aliasKey);
         }
-
+        
         if (((IClientSuggestionsProvider_Alias) source).clientcommands_isAliasSeen(aliasKey)) {
             throw RECURSIVE_ALIAS_EXCEPTION.create(aliasKey);
         }
         ((IClientSuggestionsProvider_Alias) source).clientcommands_addSeenAlias(aliasKey);
 
-        int inlineArgumentCount = (int) Pattern.compile("(?<!%)%(?:%%)*(?!%)").matcher(cmd).results().count();
-        if (inlineArgumentCount > 0) {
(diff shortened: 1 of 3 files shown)


Earthcomputer (changes requested):
Thanks for the PR! You can use [try-finally](https://docs.oracle.com/javase%2Ftutorial%2F/essential/exceptions/finally.html) in Java to make this a bit cleaner

mygizli04:
Thanks for taking the time to respond, I'm not sure if I understood exactly what you mean but I wrapped the entire function in a try-finally block. If that's not what you meant please let me know.

Earthcomputer:
This is almost good, although the try-finally should start after the addition of the alias key to the set.

mygizli04:
Alright, how about now?

src/main/java/net/earthcomputer/clientcommands/command/AliasCommand.java
```diff
@@ -78,40 +78,45 @@ private static int executeAliasCommand(FabricClientCommandSource source, String
         if (cmd == null) {
             throw NOT_FOUND_EXCEPTION.create(aliasKey);
         }
-
+
```
Earthcomputer: Can you remove this extra indentation?

Earthcomputer (changes requested):
One more thing and we're good to go.

src/main/java/net/earthcomputer/clientcommands/command/AliasCommand.java
mygizli04: Ah, my bad. Fixed now.

Earthcomputer (approved):

PR #650, opened 2024-06-24
Add /cdebug command

lugosieben:
This command opens up the debug screen (F3) for the people who do not have the F3 key on their keyboard.

You can also use `/cdebug fps|network|profiler` to show the respective graphs.

### DIFF
--- a/src/main/java/net/earthcomputer/clientcommands/command/CDebugCommand.java
+++ b/src/main/java/net/earthcomputer/clientcommands/command/CDebugCommand.java
@@ -2,29 +2,28 @@ package net.earthcomputer.clientcommands.command;
 
 import com.mojang.brigadier.Command;
 import com.mojang.brigadier.CommandDispatcher;
+import com.mojang.brigadier.context.CommandContext;
 import net.fabricmc.fabric.api.client.command.v2.FabricClientCommandSource;
 import net.minecraft.client.Minecraft;
 import net.minecraft.client.gui.components.DebugScreenOverlay;
+import net.minecraft.util.StringRepresentable;
+import org.jetbrains.annotations.NotNull;
 
+import static dev.xpple.clientarguments.arguments.CEnumArgument.*;
 import static net.fabricmc.fabric.api.client.command.v2.ClientCommandManager.*;
-import static net.earthcomputer.clientcommands.command.CDebugCommand.DebugScreenType.*;
 
 public class CDebugCommand {
     public static void register(CommandDispatcher<FabricClientCommandSource> dispatcher) {
         dispatcher.register(literal("cdebug")
-            .executes(context -> execute(OVERLAY))
-            .then(literal("overlay")
-                .executes(context -> execute(OVERLAY)))
-            .then(literal("fps")
-                .executes(context -> execute(FPS)))
-            .then(literal("network")
-                .executes(context -> execute(NETWORK)))
-            .then(literal("profiler")
-                .executes(context -> execute(PROFILER))));
+            .executes(context -> executeOverlay())
+            .then(argument("type", enumArg(DebugScreenType.class))
+                .executes(CDebugCommand::execute))
+        );
     }
 
-    private static int execute(DebugScreenType type) {
+    private static int execute(CommandContext<FabricClientCommandSource> ctx) {
         DebugScreenOverlay debugScreenOverlay = Minecraft.getInstance().getDebugOverlay();
+        DebugScreenType type = getEnum(ctx, "type");
         switch (type) {
             case OVERLAY -> debugScreenOverlay.toggleOverlay();
             case FPS -> debugScreenOverlay.toggleFpsCharts();
@@ -34,10 +33,25 @@ public class CDebugCommand {
         return Command.SINGLE_SUCCESS;
     }
     
-    enum DebugScreenType {
-        OVERLAY,
-        FPS,
-        NETWORK,
-        PROFILER
+    private static int executeOverlay() {
+        Minecraft.getInstance().getDebugOverlay().toggleOverlay();
+        return Command.SINGLE_SUCCESS;
+    }
+    
+    public enum DebugScreenType implements StringRepresentable {
+        OVERLAY("overlay"),
+        FPS("fps"),
+        NETWORK("network"),
+        PROFILER("profiler");
+        
+        private final String name;
+        
+        DebugScreenType(String name) {
+            this.name = name;
+        }
+        
+        public @NotNull String getSerializedName() {
+            return this.name;
+        }
     }
 }
--- a/src/main/java/net/earthcomputer/clientcommands/command/CDebugCommand.java
+++ b/src/main/java/net/earthcomputer/clientcommands/command/CDebugCommand.java
@@ -1,34 +1,34 @@
 package net.earthcomputer.clientcommands.command;
 
-import com.mojang.brigadier.Command;
-import com.mojang.brigadier.CommandDispatcher;
+import com.mojang.brigadier.*;
 import net.fabricmc.fabric.api.client.command.v2.FabricClientCommandSource;
 import net.minecraft.client.Minecraft;
 import net.minecraft.client.gui.components.DebugScreenOverlay;
 
-import static net.fabricmc.fabric.api.client.command.v2.ClientCommandManager.literal;
+import static net.fabricmc.fabric.api.client.command.v2.ClientCommandManager.*;
 
 public class CDebugCommand {
     public static void register(CommandDispatcher<FabricClientCommandSource> dispatcher) {
         dispatcher.register(literal("cdebug")
                 .executes((context -> execute("overlay")))
-                .then(literal("fps").executes(context -> execute("fps")))
-                .then(literal("network").executes(context -> execute("network")))
-                .then(literal("profiler").executes(context -> execute("profiler")))
+                .then(literal("overlay")
+                        .executes(context -> execute("overlay")))
+                .then(literal("fps")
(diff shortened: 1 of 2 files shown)


src/main/java/net/earthcomputer/clientcommands/command/CDebugCommand.java
```diff
@@ -0,0 +1,35 @@
+package net.earthcomputer.clientcommands.command;
+
+import com.mojang.brigadier.Command;
+import com.mojang.brigadier.CommandDispatcher;
+import net.fabricmc.fabric.api.client.command.v2.FabricClientCommandSource;
+import net.minecraft.client.Minecraft;
+import net.minecraft.client.gui.components.DebugScreenOverlay;
+
+import static net.fabricmc.fabric.api.client.command.v2.ClientCommandManager.literal;
```
xpple: Please use wildcard imports.

src/main/java/net/earthcomputer/clientcommands/command/CDebugCommand.java
```diff
@@ -0,0 +1,35 @@
...
+
+public class CDebugCommand {
+    public static void register(CommandDispatcher<FabricClientCommandSource> dispatcher) {
+        dispatcher.register(literal("cdebug")
+                .executes((context -> execute("overlay")))
+                .then(literal("fps").executes(context -> execute("fps")))
+                .then(literal("network").executes(context -> execute("network")))
+                .then(literal("profiler").executes(context -> execute("profiler")))
```
xpple: Could you format this the way other commands are formatted too? And maybe add the literal `overlay` as well; based on the suggestions people might not realise short-circuiting is an option.

src/main/java/net/earthcomputer/clientcommands/command/CDebugCommand.java
```diff
@@ -0,0 +1,35 @@
...
+        switch (type) {
+            case "overlay":
+                debugScreenOverlay.toggleOverlay(); break;
+            case "fps":
+                debugScreenOverlay.toggleFpsCharts(); break;
+            case "network":
+                debugScreenOverlay.toggleNetworkCharts(); break;
+            case "profiler":
+                debugScreenOverlay.toggleProfilerChart(); break;
+        }
```
xpple: You can use a switch expression here.

xpple (changes requested):

src/main/java/net/earthcomputer/clientcommands/command/CDebugCommand.java
```diff
@@ -1,34 +1,34 @@
 package net.earthcomputer.clientcommands.command;
 
-import com.mojang.brigadier.Command;
-import com.mojang.brigadier.CommandDispatcher;
+import com.mojang.brigadier.*;
```
xpple: I am sorry I was being ambiguous. I meant to say that you should use wildcard imports when you are importing things from DSL's like Brigadier, not everywhere. Even more so I recommend setting the threshold in your IDE for converting to wildcard imports to something like 25, the debug of 4 in IntelliJ is way too low.

Earthcomputer: Wildcards for all static imports, and for static imports only.

lugosieben: Ok, thanks for the clarification!

src/main/java/net/earthcomputer/clientcommands/command/CDebugCommand.java
```diff
@@ -0,0 +1,36 @@
...
+                .executes((context -> execute("overlay")))
+                .then(literal("overlay")
+                        .executes(context -> execute("overlay")))
+                .then(literal("fps")
+                        .executes(context -> execute("fps")))
+                .then(literal("network")
+                        .executes(context -> execute("network")))
+                .then(literal("profiler")
+                        .executes(context -> execute("profiler")))
+        );
```
xpple:
```suggestion
        dispatcher.register(literal("cdebug")
            .executes(context -> execute("overlay"))
            .then(literal("overlay")
                .executes(context -> execute("overlay")))
            .then(literal("fps")
                .executes(context -> execute("fps")))
            .then(literal("network")
                .executes(context -> execute("network")))
            .then(literal("profiler")
                .executes(context -> execute("profiler"))));
```

xpple (approved):

src/main/java/net/earthcomputer/clientcommands/command/CDebugCommand.java
```diff
@@ -0,0 +1,35 @@
...
+        dispatcher.register(literal("cdebug")
+            .executes(context -> execute("overlay"))
+            .then(literal("overlay")
+                .executes(context -> execute("overlay")))
+            .then(literal("fps")
+                .executes(context -> execute("fps")))
+            .then(literal("network")
+                .executes(context -> execute("network")))
+            .then(literal("profiler")
+                .executes(context -> execute("profiler"))));
```
RealRTTV: nit: Use a `CEnumArgument` instead of string literals

RealRTTV (changes requested):
cool idea, an enum would be more future-proof than string literals, and less code to type for the branches

RealRTTV:
why can't a CEnumArgument be used here?

lugosieben:
> why can't a CEnumArgument be used here?

sorry, i just noticed that you wrote 2 comments and I wasn't aware of the CEnumArgument.

lugosieben:
Sorry to bother, but what's the status on this PR?

xpple:
It is currently open.

lugosieben:
Yeah, but I mean if it's going to be merged sometime, if the feature seems unnecessary and not getting added or if I need to update something on the code. Sry for not clarifying. If you don't know yet, it's also ok, just wanted to ask

src/main/java/net/earthcomputer/clientcommands/command/CDebugCommand.java
```diff
@@ -0,0 +1,57 @@
...
+        dispatcher.register(literal("cdebug")
+            .executes(context -> executeOverlay())
+            .then(argument("type", enumArg(DebugScreenType.class))
+                .executes(CDebugCommand::execute))
+        );
+    }
+
+    private static int execute(CommandContext<FabricClientCommandSource> ctx) {
+        DebugScreenOverlay debugScreenOverlay = Minecraft.getInstance().getDebugOverlay();
+        DebugScreenType type = getEnum(ctx, "type");
```
xpple: Please make this method take the command source and enum as parameters. That way you can do `source#getClient` instead of `Minecraft#getInstance` and make it consistent with the other commands.

src/main/java/net/earthcomputer/clientcommands/command/CDebugCommand.java
```diff
@@ -0,0 +1,57 @@
...
+            case NETWORK -> debugScreenOverlay.toggleNetworkCharts();
+            case PROFILER -> debugScreenOverlay.toggleProfilerChart();
+        }
+        return Command.SINGLE_SUCCESS;
+    }
+    
+    private static int executeOverlay() {
+        Minecraft.getInstance().getDebugOverlay().toggleOverlay();
+        return Command.SINGLE_SUCCESS;
+    }
```
xpple: Same applies to this method.

xpple (changes requested):
Looks good, one final thing.

src/main/java/net/earthcomputer/clientcommands/command/CDebugCommand.java
```diff
@@ -0,0 +1,49 @@
...
+        switch (type) {
+            case OVERLAY -> debugScreenOverlay.toggleOverlay();
+            case FPS -> debugScreenOverlay.toggleFpsCharts();
+            case NETWORK -> debugScreenOverlay.toggleNetworkCharts();
+            case PROFILER -> debugScreenOverlay.toggleProfilerChart();
+        }
+        return Command.SINGLE_SUCCESS;
+    }
+
+    public enum DebugScreenType implements StringRepresentable {
```
xpple: Final, final thing, the enum can be private.

src/main/java/net/earthcomputer/clientcommands/command/CDebugCommand.java
```diff
@@ -0,0 +1,49 @@
...
+        );
+    }
+
+    private static int execute(FabricClientCommandSource source, DebugScreenType type) {
+        DebugScreenOverlay debugScreenOverlay = source.getClient().getDebugOverlay();
+        switch (type) {
+            case OVERLAY -> debugScreenOverlay.toggleOverlay();
+            case FPS -> debugScreenOverlay.toggleFpsCharts();
+            case NETWORK -> debugScreenOverlay.toggleNetworkCharts();
+            case PROFILER -> debugScreenOverlay.toggleProfilerChart();
```
RealRTTV, line 28: I'd move over the toggle to be a method called `toggle` in the enum, since it is an interaction that's related to the enum rather than the interaction of the command on the enum

RealRTTV (changes requested):
Looks good, one final thing.

src/main/java/net/earthcomputer/clientcommands/command/CDebugCommand.java
xpple: I think that's a bit pedantic, it's fine now.

xpple (approved):

xpple:
Thank you! For such a simple command, it went through quite a lot of reviewing 😄.

PR #659, opened 2024-09-01
Add glow argument to find command.

electricsteve:
Add a glow argument to the find command to glow the found entity/entities.
Discussion points:
- Time for glow
- Color of glow
Should these be changeable by the player?

For anyone wondering why:
- Easily locate the entities
- When using effect to glow all entities you want, it doesn't apply to those outside of render distance, with keep searching this is easier


Earthcomputer:
Isn't this the same as `/cglow entities`?

electricsteve:
🤦🏼

Issue #574, opened 2023-10-31
/cghostentity


Cookie499:
I want to use the command like /cghostentity to create a boat only exist in client, which allow player match any coordinates accurately by lean anginst the boats.

Earthcomputer:
For your exact issue, we could possibly also add a `/csnap` command to teleport to specific coordinates less than 1 block away.

Wikinger8:
> For your exact issue, we could possibly also add a `/csnap` command to teleport to specific coordinates less than 1 block away.

This function would be so cool. Should I create a new issue for it?

Earthcomputer:
Yes

PR #661, opened 2024-09-16
Add a simple csnap command

xpple:
This commands allows you to snap to any location within a unit sphere around the player's feet. Passing no arguments enables click to teleporting, allowing the player to snap to the location their crosshair is pointed at. Otherwise passing coordinates teleports the player to said coordinates, with the above restriction.

The idea for this command was mentioned in #574.


Earthcomputer:
I made some improvements to this PR so that it can be merged now.
Aside from code style:
- It now checks whether the player can fit at the target location in the current pose
- Click-to-snap now only triggers on a right click, rather than every 4 ticks while the right mouse is held down
- Click-to-snap now only works when you click on the top face of a block
- Click-to-snap doesn't work when the player is airborne, to prevent an easy way to cheat at parkour. Of course, they can still type in the coordinates manually but at least it's not quite as trivial.

PR #663, opened 2024-09-26
Simplify c2c packets

xpple:
Big thanks to @wagyourtail. Since we can rely on the `sender` field of a `ClientboundPlayerChatPacket`, it is superfluous to include this information in the C2C packet too. This PR excludes that field from transmission.

Because of how this is implemented, listening to C2C packets through `/clisten` will display the `sender` field as if it were encoded in the C2C packet. On the one hand this makes sense; the C2C packet classes all have this field. On the other hand it doesn't; it is not actually included in the packet that is transmitted. I'll let others comment on which is better. It could even be made configurable.


src/main/java/net/earthcomputer/clientcommands/c2c/RawPacketInfo.java
```diff
@@ -0,0 +1,18 @@
+package net.earthcomputer.clientcommands.c2c;
+
+import io.netty.buffer.ByteBuf;
+import net.minecraft.core.RegistryAccess;
+import net.minecraft.network.RegistryFriendlyByteBuf;
+
+public class RawPacketInfo extends RegistryFriendlyByteBuf {
```
Earthcomputer: Maybe `C2CFriendlyByteBuf`?

Earthcomputer (approved):

xpple:
The question I posed remains open.

Earthcomputer:
I took a look at the code and I thought it was fine, I don't think the decision is super consequential for now. It can always be changed later

PR #669, opened 2024-10-28
Update clientarguments for 1.21.3

xpple:
PR for the forthcoming release of clientarguments.


src/main/resources/fabric.mod.json
```diff
@@ -29,7 +29,8 @@
     "minecraft": "${mcversion}",
     "fabricloader": ">=0.4.0",
     "fabric-api": "*",
-    "clientarguments": ">=1.3.0"
+    "clientarguments": ">=1.10.0",
```
Earthcomputer: I think there's no reason for these dependencies since they're already JiJ'ed. Let's just remove them

PR #698, opened 2024-12-23
Add the cserverseed command

xpple:
This PR adds a utility command for inspecting two seeds the server sends you for free:

- The lower 16 bits of the enchantment seed
- The lower 8 bytes of the SHA256-hashed world seed

In combination with the structure seed, the lower eight bytes of the world seed hash uniquely pinpoint the world seed. I added a subcommand for this too.


Johnsmith927:
How is this better than the current method of seedcracking? In both methods the player needs to crack the structure seed, no? I could be wrong, I'm just curious

xpple:
This command isn’t seed cracking in any way. It’s just for inspecting two bits of information the server sends the player for free.

The subcommand that determines the world seed from the structure seed and the hashed seed is just a brute force over all remaining world seed possibilities.

PR #704, opened 2025-01-02
Add minimal Checkstyle configuration

xpple:
This PR adds a minimal Checkstyle based on the [code style guidelines](https://github.com/Earthcomputer/clientcommands/blob/8a73f39ef032d84603d95a1b77e74d678182e424/docs/CONTRIBUTING.md#code-style).


checkstyle.xml
```diff
@@ -0,0 +1,29 @@
...
+<module name="Checker">
+    <property name="fileExtensions" value="java"/>
+    <module name="NewlineAtEndOfFile"/>
+
+    <module name="TreeWalker">
+        <module name="SuppressionCommentFilter"/>
+
+        <module name="AvoidStarImport">
+            <property name="allowStaticMemberImports" value="true"/>
+            <property name="excludes" value="net.earthcomputer.clientcommands.command"/>
```
Earthcomputer: What's this line for?

xpple: It is to exclude the wildcard import in `ClientCommands.java` from the check.

Earthcomputer: Okay then maybe suppress that one with a comment rather than globally in the config

xpple: Done.

Earthcomputer (approved):
Still don't like checkstyle but whatever it's better to catch some of the more obvious things in PRs before they get to us.

PR #706, opened 2025-01-07
Concatenate sign lines to allow for multiline search

### LINKED ISSUE #705: How do I use /csignsearch to find signs with required text being across multiple lines?
Let's say I have a lot of signs with different values on the first two lines:
```
20
100

80
60

100
70

30
80

90
90

50
20
```

And I need to find specifically the "30, 80" sign.
I am very new to regex, so the only idea I had was to do `30\n80`, which failed.
Also tried `30.*80` since sign data is stored as single line, but same result.
How can I do this?

I am on v2.9, but changelog doesn't show any changes to this commands, so it probably shouldn't be different from latest version.

xpple:
Closes #705.

### DIFF
--- a/src/main/java/net/earthcomputer/clientcommands/command/SignSearchCommand.java
+++ b/src/main/java/net/earthcomputer/clientcommands/command/SignSearchCommand.java
@@ -18,6 +18,8 @@ import net.minecraft.world.level.block.entity.SignText;
 import net.minecraft.world.level.block.state.BlockState;
 
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
 
 import static com.mojang.brigadier.arguments.StringArgumentType.*;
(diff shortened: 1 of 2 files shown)


Earthcomputer:
I was thinking that the regex argument type and possible the regexes themselves would also need to be adjusted to support multiple lines.

xpple:
Assuming you are referring to modifiers like `g` and `m`: `g` is pointless because we are looking for any match and `m` already works.

xpple:
Upon closer inspection `m` does not in fact work 😄.

xpple:
Should work now.

Earthcomputer (changes requested):
There's still no way to input a newline character in the regex argument type

xpple:
That's just `\n`.

Earthcomputer (approved):
So it does 😂

PR #712, opened 2025-02-04
Add the ctick command

xpple:
This PR adds a simple `ctick` command for querying both the client's and server's TPS and MSPT. The expected value is printed for completeness. The task code is mostly copied from:

https://github.com/Earthcomputer/clientcommands/blob/c1e37665739f0e1d6aeb826b9d4e45b7adb5d876/src/main/java/net/earthcomputer/clientcommands/command/CommandTick.java#L175-L255

### DIFF
--- a/src/main/java/net/earthcomputer/clientcommands/command/CTickCommand.java
+++ b/src/main/java/net/earthcomputer/clientcommands/command/CTickCommand.java
@@ -161,7 +161,7 @@ public class CTickCommand {
                 long totalTime = lastTickStart - firstTickStart;
                 double mspt = totalTime / (1_000_000D * tickCount);
                 ClientCommandHelper.sendFeedback(Component.translatable("commands.ctick.mspt", DEC_FMT.format(mspt)));
-                ClientCommandHelper.sendFeedback(Component.translatable("commands.ctick.mspt.inaccurate"));
+                ClientCommandHelper.sendHelp(Component.translatable("commands.ctick.mspt.inaccurate"));
             } else {
                 double mspt = totalTickTime / (1_000_000D * tickCount);
                 ClientCommandHelper.sendFeedback(Component.translatable("commands.ctick.mspt", DEC_FMT.format(mspt)));
--- a/src/main/java/net/earthcomputer/clientcommands/command/CTickCommand.java
+++ b/src/main/java/net/earthcomputer/clientcommands/command/CTickCommand.java
@@ -22,7 +22,7 @@ public class CTickCommand {
 
     private static final String TASK_NAME = "ctick";
 
-    private static TickMeasuringTask currentMeasurer = null;
+    private static CurrentTask currentTask = null;
 
     public static void register(CommandDispatcher<FabricClientCommandSource> dispatcher) {
         dispatcher.register(literal("ctick")
@@ -42,8 +42,8 @@ public class CTickCommand {
         stopPreviousTask();
 
         TickMeasuringTask measurer = new TickMeasuringTask(true, false);
-        TaskManager.addTask(TASK_NAME, measurer);
-        currentMeasurer = measurer;
+        String name = TaskManager.addTask(TASK_NAME, measurer);
+        currentTask = new CurrentTask(name, measurer);
 
         float tps = source.getWorld().tickRateManager().tickrate();
         source.sendFeedback(Component.translatable("commands.ctick.client.tps.expectedTps", tps));
@@ -54,8 +54,8 @@ public class CTickCommand {
         stopPreviousTask();
 
         TickMeasuringTask measurer = new TickMeasuringTask(false, false);
-        TaskManager.addTask(TASK_NAME, measurer);
-        currentMeasurer = measurer;
+        String name = TaskManager.addTask(TASK_NAME, measurer);
+        currentTask = new CurrentTask(name, measurer);
 
         float mspt = TimeUtil.MILLISECONDS_PER_SECOND / source.getWorld().tickRateManager().tickrate();
         source.sendFeedback(Component.translatable("commands.ctick.client.mspt.expectedMspt", mspt));
@@ -67,8 +67,8 @@ public class CTickCommand {
 
         boolean isIntegratedServer = source.getClient().hasSingleplayerServer();
         TickMeasuringTask measurer = new TickMeasuringTask(true, !isIntegratedServer);
-        TaskManager.addTask(TASK_NAME, measurer);
-        currentMeasurer = measurer;
+        String name = TaskManager.addTask(TASK_NAME, measurer);
+        currentTask = new CurrentTask(name, measurer);
 
         float tps = source.getWorld().tickRateManager().tickrate();
         source.sendFeedback(Component.translatable("commands.ctick.server.tps.expectedTps", tps));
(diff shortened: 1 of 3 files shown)


src/main/java/net/earthcomputer/clientcommands/command/CTickCommand.java
```diff
@@ -0,0 +1,210 @@
...
+
+    private static void stopPreviousTask() {
+        if (currentMeasurer != null) {
+            currentMeasurer._break();
+            TaskManager.removeTask(TASK_NAME);
+            currentMeasurer = null;
+        }
+    }
+
+    // see https://github.com/Earthcomputer/clientcommands/blob/c1e37665739f0e1d6aeb826b9d4e45b7adb5d876/src/main/java/net/earthcomputer/clientcommands/command/CommandTick.java#L175-L255
```
Earthcomputer: Don't think we need this comment

src/main/java/net/earthcomputer/clientcommands/command/CTickCommand.java
```diff
@@ -0,0 +1,210 @@
...
+
+        float mspt = TimeUtil.MILLISECONDS_PER_SECOND / source.getWorld().tickRateManager().tickrate();
+        source.sendFeedback(Component.translatable("commands.ctick.server.mspt.expectedMspt", mspt));
+        return (int) mspt;
+    }
+
+    private static void stopPreviousTask() {
+        if (currentMeasurer != null) {
+            currentMeasurer._break();
+            TaskManager.removeTask(TASK_NAME);
```
Earthcomputer: This is not how tasks work. The string argument of `addTask` is a template which only forms part of the task's full name, which `addTask` returns. This therefore won't remove the task properly. You also don't need to call `_break` because `removeTask` does that for you.

src/main/java/net/earthcomputer/clientcommands/command/CTickCommand.java
```diff
@@ -0,0 +1,210 @@
...
+        protected void onTick() {
+            if (tickCount >= PERIOD) {
+                lastTickStart = System.nanoTime();
+                _break();
+            }
+        }
+
+        @Override
+        public void initialize() {
+            minecraft.player.displayClientMessage(Component.translatable("commands.ctick.measuring"), false);
```
Earthcomputer: All these can be replaced with `ClientCommandHelper.sendFeedback`

src/main/java/net/earthcomputer/clientcommands/command/CTickCommand.java
```diff
@@ -0,0 +1,210 @@
...
+                minecraft.player.displayClientMessage(Component.translatable("commands.ctick.mspt.inaccurate"), false);
+            } else {
+                double mspt = totalTickTime / (1_000_000D * tickCount);
+                minecraft.player.displayClientMessage(Component.translatable("commands.ctick.mspt", DEC_FMT.format(mspt)), false);
+            }
+        }
+
+        @Override
+        public boolean condition() {
+            return tickCount <= 1200;
```
Earthcomputer, line 172: This wasn't a problem in 1.12 since the client tick rate was always expected to be 20 and hardly ever fell below 20, but now if the TPS is set to 1 then this could take a long time to complete. I would suggest using real time instead. Just be extra careful about there being zero ticks elapsed and dividing by zero.

src/main/resources/assets/clientcommands/lang/en_us.json
```diff
@@ -249,6 +249,16 @@
   "commands.ctask.stop.noMatch": "No matching tasks",
   "commands.ctask.stop.success": "Stopped %s tasks",
 
+  "commands.ctick.client.mspt.expectedMspt": "The expected client MSPT is %s",
+  "commands.ctick.client.tps.expectedTps": "The expected client TPS is %s",
+  "commands.ctick.measuring": "Measuring...",
+  "commands.ctick.mspt": "Milliseconds per tick = %s",
+  "commands.ctick.mspt.inaccurate": "MSPT is inaccurate above 20TPS",
```
Earthcomputer: The TPS isn't necessarily 20 anymore.
```suggestion
  "commands.ctick.mspt.inaccurate": "MSPT is inaccurate when the server isn't lagging",
```

src/main/java/net/earthcomputer/clientcommands/command/CTickCommand.java
```diff
@@ -0,0 +1,210 @@
...
+
+    private static int getClientTps(FabricClientCommandSource source) throws CommandSyntaxException {
+        stopPreviousTask();
+
+        TickMeasuringTask measurer = new TickMeasuringTask(true, false);
+        TaskManager.addTask(TASK_NAME, measurer);
+        currentMeasurer = measurer;
+
+        float tps = source.getWorld().tickRateManager().tickrate();
+        source.sendFeedback(Component.translatable("commands.ctick.client.tps.expectedTps", tps));
```
Earthcomputer, line 48: These expected TPS and MSPT messages shouldn't be printed while the game is sprinting

src/main/java/net/earthcomputer/clientcommands/command/CTickCommand.java
```diff
@@ -0,0 +1,210 @@
...
+        float tps = source.getWorld().tickRateManager().tickrate();
+        source.sendFeedback(Component.translatable("commands.ctick.server.tps.expectedTps", tps));
+        return (int) tps;
+    }
+
+    private static int getServerMspt(FabricClientCommandSource source) throws CommandSyntaxException {
+        stopPreviousTask();
+
+        boolean isIntegratedServer = source.getClient().hasSingleplayerServer();
+        TickMeasuringTask measurer = new TickMeasuringTask(false, !isIntegratedServer);
```
Earthcomputer, line 81: I'm not seeing where there is any accurate measurements from the integrated server. Be careful about thread safety with that, though.

src/main/java/net/earthcomputer/clientcommands/command/CTickCommand.java
```diff
@@ -0,0 +1,210 @@
...
+        public void onCompleted() {
+            if (tps) {
+                long totalTime = lastTickStart - firstTickStart;
+                double tps = 1_000_000_000D * tickCount / totalTime;
+                minecraft.player.displayClientMessage(Component.translatable("commands.ctick.tps", totalTime == 0 ? Component.translatable("commands.ctick.tps.immeasurable") : DEC_FMT.format(tps)), false);
+            } else if (forceInaccurate) {
+                long totalTime = lastTickStart - firstTickStart;
+                double mspt = totalTime / (1_000_000D * tickCount);
+                minecraft.player.displayClientMessage(Component.translatable("commands.ctick.mspt", DEC_FMT.format(mspt)), false);
+                minecraft.player.displayClientMessage(Component.translatable("commands.ctick.mspt.inaccurate"), false);
```
Earthcomputer: This message shouldn't display if the server is clearly lagging (e.g. the observed TPS is say 10% lower than expected). It also shouldn't display if the game is sprinting. Also for cosmetic reasons maybe it could be replaced with one of those help comments (`ClientCommandHelper.sendHelp`)

Earthcomputer (changes requested):

src/main/java/net/earthcomputer/clientcommands/command/CTickCommand.java
xpple: This was copied over from your original code which I wanted to stay as close to as possible.

Earthcomputer: The concept of help messages didn't exist back then, and besides we can improve on code I wrote 7 years ago :)

src/main/java/net/earthcomputer/clientcommands/command/CTickCommand.java
xpple: I added it to inform readers of the code that the code is originally from somewhere else.

Earthcomputer: It's the same project, doesn't need credit

xpple: I was more thinking about the author, but I'll remove it.

PR #725, opened 2025-03-24
Add cfps command

### LINKED ISSUE #711: Ability to set maximum framerate via command
Ability to set maximum framerate via command, since in minecraft the highest limit you can set is 250 FPS, I would like for example set it to 280 fps because my monitor supports 280 Hz at max wich I use.

xpple:
Closes #711. The unlimited framerate cutoff is now increased to the maximum of the refresh rates of all connected monitors. This allows people to set their maximum framerate to a value possibly higher than `260`, without effectively setting it to unlimited.

### DIFF
--- a/src/main/java/net/earthcomputer/clientcommands/command/FramerateCommand.java
+++ b/src/main/java/net/earthcomputer/clientcommands/command/FramerateCommand.java
@@ -1,0 +1,29 @@
+package net.earthcomputer.clientcommands.command;
+
+import com.mojang.brigadier.CommandDispatcher;
+import net.fabricmc.fabric.api.client.command.v2.FabricClientCommandSource;
+import net.minecraft.client.Minecraft;
+import net.minecraft.network.chat.Component;
+
+import static com.mojang.brigadier.arguments.IntegerArgumentType.*;
+import static net.fabricmc.fabric.api.client.command.v2.ClientCommandManager.*;
+
+public class FramerateCommand {
+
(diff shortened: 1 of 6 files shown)


src/main/java/net/earthcomputer/clientcommands/command/FramerateCommand.java
```diff
@@ -0,0 +1,45 @@
...
+import net.minecraft.network.chat.Component;
+
+import java.util.function.IntSupplier;
+
+import static com.mojang.brigadier.arguments.IntegerArgumentType.*;
+import static net.fabricmc.fabric.api.client.command.v2.ClientCommandManager.*;
+
+public class FramerateCommand {
+
+    public static final IntSupplier MAX_REFRESH_RATE = () -> Minecraft.getInstance().virtualScreen.screenManager.monitors.values().stream()
```
Earthcomputer: Why not make this a method?

xpple: I intended to make it come across as a constant that only rarely changes.

Earthcomputer: That's not a good reason to not use a method

xpple: What's wrong with a supplier though?

xpple:
Closed in favour of #726.

PR #734, opened 2025-05-05
Move rendering to after translucent

xpple:
This fixes the cuboids being drawn behind things like water even though they are physically in front.


Earthcomputer (approved):
Seems way more consistent than it used to be, good job!

PR #739, opened 2025-06-14
Update zh_cn.json

GodGun968:


Earthcomputer (changes requested):
Hi, there are some errors you need to fix:
```
 zh_cn.json:413: translation key 'twoPlayerGame.chat.won' does not have enough format specifiers. It only has 1 while the English has 2
zh_cn.json:413: translation key 'twoPlayerGame.chat.won' contains an unescaped '%'. Try escaping it with '%%'
zh_cn.json:415: translation key 'twoPlayerGame.incoming' does not have enough format specifiers. It only has 1 while the English has 2
zh_cn.json:415: translation key 'twoPlayerGame.incoming' contains an unescaped '%'. Try escaping it with '%%'
```
looks like you accidentally broke some `%s` format specifiers on line 413 and line 415.

Earthcomputer:
See the [translation guideline](https://github.com/Earthcomputer/clientcommands/blob/fabric/docs/TRANSLATING.md#string-substitutions) on string substitutions if you're confused

GodGun968:
Sorry, this is because there is similar content so I copied it directly but it was copied to the wrong location

GodGun968:
> [translation guideline](https://github.com/Earthcomputer/clientcommands/blob/fabric/docs/TRANSLATING.md#string-substitutions)

And the "use %1$s, %2$s, %3$s" is because the text was separated at the time and I had to translate it into such a word order. I actually prefer the word order of the original text

GodGun968:
And can you explain what "RNG call" means and what does "callback" do? It is best to have some demo video

Earthcomputer:
> can you explain what "RNG call" means

An "RNG call" in this context is a usage of the player RNG. Unexpected usages of the player RNG can mess up RNG manipulation because you're now in the wrong part of the RNG number sequence and will get different results.

> what does "callback" do

`/ccallback` is a sort-of internal command to enable clientcommands to run code when you click on buttons in chat. The callback error will likely happen if the user tries to run the command directly without clientcommands having registered the callback. Microsoft seems to translate callback as "回拨", see their [terminology search website](https://learn.microsoft.com/en-us/globalization/reference/microsoft-terminology)

GodGun968:
> `/ccallback` is a sort-of internal command to enable clientcommands to run code when you click on buttons in chat. The callback error will likely happen if the user tries to run the command directly without clientcommands having registered the callback.

well, I probably know how to translate it, but is there a demo video? what effect will it show?

ccallback -> 回调 RNG back -> RNG调用

Do you think this translation is appropriate? I remember there is such a thing in programming.

Earthcomputer:
The callback has nothing to do with RNG, it's separate. It's a callback in the programming sense. Chatgpt says that how you have just translated it is good :+1:

Earthcomputer (approved):

GodGun968:
[call](https://baike.baidu.com/item/%E8%B0%83%E7%94%A8/3351318) I didn't find it in the Wikipedia

[callback](https://en.wikipedia.org/wiki/Callback_(computer_programming))

Is this explanation correct?

Earthcomputer:
callback is correct. As for RNG call, I'm unable to provide much input on whether that's correct as I don't speak Chinese, but here's some advice from ChatGPT, feel free to ignore it in case it's speaking rubbish:
image

GodGun968:
OK Thank you for your review

> The callback has nothing to do with RNG, it's separate. It's a callback in the programming sense. Chatgpt says that how you have just translated it is good 👍

GodGun968:
A piece of code is used to modify RNG. So because some behaviors in the game cause the modified RNG program to be called incorrectly, causing the RNG to change, so cannot be cracked. 
Is this explanation correct?

Earthcomputer:
Not really. Think of the RNG as a function which supplies random numbers. So an RNG call is a call (aka invocation) to this function. It's a function call, just like this:
image

GodGun968:
Oh, sorry I'm confused, so RNG itself is a piece of code, right?

Earthcomputer:
Yes, exactly, it's a piece of code which generates random numbers. And when you're doing RNG manipulation, you don't want to call that piece of code when you're not supposed to, and that is what this warning/error message is about. The warning says that the RNG function has been called when it wasn't supposed to have been called, and then `%s` is the thing that the player did to cause the RNG to have been called (e.g. sprinting).

GodGun968:
OK, just like I understand, so there is no problem with this translation

PR #723, opened 2025-03-20
Add some improvements to cwaypoint

xpple:
Closes #721.

PR #749, opened 2025-08-13
Migrate to SimpleWaypoints

### LINKED ISSUE #721: Waypoint improvements
A few enhancement ideas for the `/cwaypoint` command:

### Show/Hide waypoints
The `/cwaypoint show <name>` and `/cwaypoint hide <name>` commands to show or hide a waypoint.

### Set waypoint color
The ability to set a color for the waypoints, instead of the default yellow color. It can be:
- `/cwaypoint color <name> <hex color>`
- `/cwaypoint add <name> <pos> [<dimension>] [<color>]`

### Auto hide waypoints
A config option like `/cconfig clientcommands hideWaypointDistance set <value>` with a default value of 0. If the player is closer than this value to a waypoint, it will disappear/fade, and if the player gets further, it will reappear again.

xpple:
See [https://github.com/xpple/SimpleWaypoints](https://github.com/xpple/SimpleWaypoints). Closes #721, supersedes #723.


src/main/java/net/earthcomputer/clientcommands/ClientCommands.java
```diff
@@ -82,7 +88,53 @@ public void onInitializeClient() {
         FishingCracker.registerEvents();
         PlayerRandCracker.registerEvents();
         ServerBrandManager.registerEvents();
-        WaypointCommand.registerEvents();
+    }
+
+    private static void migrateWaypoints() {
```
Earthcomputer: I think this should go in a different class. Even though it's only one function it's big enough that it could go in a class on its own, maybe in `features.Waypoints` or something.

build.gradle
```diff
@@ -61,6 +61,7 @@ dependencies {
 
 	include modImplementation("dev.xpple:clientarguments:${project.clientarguments_version}")
 	include modImplementation("dev.xpple:betterconfig-fabric:${project.betterconfig_version}")
+	include modImplementation("dev.xpple:simplewaypoints:${project.simplewaypoints_version}")
```
Earthcomputer, line 64: Is simplewaypoints a project on modrinth? You could add it as an included library in the `modrinth {}` block below.

src/main/java/net/earthcomputer/clientcommands/mixin/simplewaypoints/WaypointCommandMixin.java
```diff
@@ -0,0 +1,21 @@
...
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+
+import static net.fabricmc.fabric.api.client.command.v2.ClientCommandManager.*;
+
+@Mixin(WaypointCommand.class)
+public class WaypointCommandMixin {
+    @Inject(method = "register", at = @At("TAIL"), remap = false)
+    private static void registerWaypointCommandAlias(CommandDispatcher<FabricClientCommandSource> dispatcher, CallbackInfo ci) {
+        CommandNode<FabricClientCommandSource> configRoot = dispatcher.getRoot().getChild("sw:waypoint");
+        dispatcher.register(literal("cwaypoint").redirect(configRoot));
```
Earthcomputer: I don't think this needs to be a mixin. There are a few other ways of doing this, maybe cleanest would be to add an API to simplewaypoints for aliases.

PR #752, opened 2025-08-24
Add the cpredictbrushables command

### LINKED ISSUE #743: Predict suspicious gravel drops
pretty self explanatory but if say you had the seed and the coordinates of the gravel block would it be possible to predict the loot item it will drop, 
or would it be similar to trial vaults where it depends on what tick u start brushing the block so it has to be predicted via other methods

xpple:
Depends on [SeedFinding/mc_feature_java#10](https://github.com/SeedFinding/mc_feature_java/pull/10). Closes #743.


src/main/java/net/earthcomputer/clientcommands/command/PredictBrushablesCommand.java
```diff
@@ -0,0 +1,152 @@
...
+    public static void register(CommandDispatcher<FabricClientCommandSource> dispatcher) {
+        var cpredictbrushables = dispatcher.register(literal("cpredictbrushables")
+            .executes(ctx -> predictBrushables(ctx, Component.translatable("commands.cpredictbrushables.starting"))));
+        FLAG_KEEP_SEARCHING.addToCommand(dispatcher, cpredictbrushables, ctx -> true);
+    }
+
+    public static int predictBrushables(CommandContext<FabricClientCommandSource> ctx, Component startingMessage) throws CommandSyntaxException {
+        boolean keepSearching = getFlag(ctx, FLAG_KEEP_SEARCHING);
+        sendFeedback(startingMessage);
+        TaskManager.addTask("cpredictbrushables", new PredictBrushablesTask(keepSearching));
```
Earthcomputer: If `keepSearching` is true, then maybe there should be a way to cancel it.

xpple: `/ctask stop`?

Earthcomputer: Yeah, the task name that can be cancelled is returned by the `addTask` method.

xpple: It wasn't a question, I meant to say that there already is a way to cancel the task.

Earthcomputer: Yeah but having a cancel button that runs that commands for you, like other long running tasks already have, is better UX

xpple: What do you mean exactly? E.g. `/cfindblock --keep-searching` has no such button when you run it. Or I can add it there as well?

PR #761, opened 2025-10-31
Enchantment prediction speedup

z7087:
I conducted a performance comparison between my custom build and the official build.

Test Parameters:

Command: /cenchant minecraft:diamond_boots with minecraft:feather_falling 4 with minecraft:protection 4
Configuration: maxEnchantItemThrows=50000
Results:

My Custom Build: 3-5 seconds
Official Build: 10+ seconds
I have a couple of questions regarding potential optimizations:

In a loop, can RandomSource.create() be safely replaced with RandomSource.createNewThreadLocalInstance()?
When I increased the test parameters to maxEnchantItemThrows=100000, my build experienced significant FPS spikes, and the task took over 15 seconds to complete. Should I consider reducing the thread pool size to address this?

### DIFF
--- a/src/main/java/net/earthcomputer/clientcommands/features/EnchantmentCracker.java
+++ b/src/main/java/net/earthcomputer/clientcommands/features/EnchantmentCracker.java
@@ -115,6 +115,8 @@ public class EnchantmentCracker {
 
     public static final Logger LOGGER = LogUtils.getLogger();
     private static final int PROGRESS_BAR_WIDTH = 50;
+    private static final LCG LCG_JAVA_ADVANCE_4 = LCG.JAVA.combine(4);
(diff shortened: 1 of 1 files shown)


z7087:
My CPU is a 4-core, 8-thread model, and Runtime.getRuntime().availableProcessors() returns 8. After reducing the thread pool size by approximately half for both builds, I recorded times of 9s and 13s respectively, and the FPS drops were eliminated.

Earthcomputer:
@z7087 Why did you close the PR?

z7087:
> @z7087 Why did you close the PR?

I closed the PR because I realized the root cause of the FPS drops was an oversized thread pool, not the calculation logic itself. My optimization doesn't significantly reduce computation time, so its impact is limited. A better approach might be to make the thread pool size a configurable option instead.

PR #762, opened 2025-11-15
fix crash on saving overworld.dat

GuiTaek:
For me saving e.g. overworld.dat with -Dclientcommands.debugDimensionRng=overworld as I was told on quitting a world crashes without this fix on windows 11, because windows doesn't allow colon (":") in file paths and .toString gives "minecraft:overworld"


Earthcomputer:
Changed it a little and merged manually, thanks!

PR #775, opened 2026-02-09
Improve the /cwiki command

makaron-24:
Added `[Open article]` clickable message sent after the article content and searching, mainly for accepting typos when entering the page name.

### DIFF
--- a/src/main/java/net/earthcomputer/clientcommands/command/WikiCommand.java
+++ b/src/main/java/net/earthcomputer/clientcommands/command/WikiCommand.java
@@ -8,6 +8,7 @@ import net.fabricmc.fabric.api.client.command.v2.FabricClientCommandSource;
 import net.minecraft.ChatFormatting;
 import net.minecraft.network.chat.ClickEvent;
 import net.minecraft.network.chat.Component;
+import org.jspecify.annotations.Nullable;
 
 import java.net.URI;
 import java.net.URLEncoder;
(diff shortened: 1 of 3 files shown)


src/main/java/net/earthcomputer/clientcommands/command/WikiCommand.java
```diff
@@ -5,23 +5,54 @@
...
     }
 
+    @Nullable
+    private static Component getLinkComponent(String page) {
+        String title = WikiRetriever.searchArticleName(page);
+        if (title == null) {
+            return null;
+        }
+
+        String pageName = URLEncoder.encode(title, StandardCharsets.UTF_8).replace('+', '_');
```
Earthcomputer, line 40: I'm not sure I understand this line. Is it meant to be replacing spaces with underscores? Spaces get replaced with `%20` after URL encoding, not `+`. Alternatively you could replace them before URL encoding, as I'm not sure that `_` gets changed with URL encoding (not 100% sure on this)?

src/main/java/net/earthcomputer/clientcommands/features/WikiRetriever.java
```diff
@@ -184,19 +188,78 @@ public static String decode(String html) {
...
+            return null;
+        }
+        return result;
+    }
+
+    @Nullable
+    public static URL buildURL(String page, String query) {
+        String result = Arrays.stream(page.split("\\s+"))
+                .map(w -> w.substring(0, 1).toUpperCase()
+                        + w.substring(1).toLowerCase())
```
Earthcomputer: Every time you call `toUpperCase` or `toLowerCase`, you need to pass in `Locale.ROOT` otherwise it uses the current system locale (unless this is what you intended). Notably in Turkish the i with and without a dot are two separate letters, and so capitalizing i doesn't produce what you would expect.

Earthcomputer (changes requested):

src/main/java/net/earthcomputer/clientcommands/command/WikiCommand.java
makaron-24: Well, it might be outdated but this documentation says otherwise: https://docs.oracle.com/javase/8/docs/api/java/net/URLEncoder.html

xpple: Not outdated, it works the same on Java 21.

Earthcomputer (approved):

PR #790, opened 2026-05-17
fix #395

### LINKED ISSUE #395: Ambiguity between arguments in `/cfind`
Starting up Minecraft with Client Commands installed prints the following warning to the game log:
```
Ambiguity between arguments [cfind, --keep-searching] and [cfind, filter] with inputs: [--keep-searching]
```

I don't recall this happening on Client Commands versions earlier than 2.6.5, but it does happen in that version and later.
Minecraft 1.18.2
Fabric API 0.47.8

RealRTTV:
not only does this have a log message but has a noticeable impact on syntax-highlighting with the ambiguity removed


RealRTTV:
also for consistency should we also add exclusions for all flags, or just the ones where the parser accepts anything so it *needs* an exclusion?

Earthcomputer (approved):

PR #791, opened 2026-05-17
fix #568

### LINKED ISSUE #568: Various games oversized on gui scale auto
With horizontal window sizes and gui scale set to auto, the top and bottom of the snake game screen appears outside of the view frame

With borderless fullscreen, 1920x1080:

With border fullscreen, 1920x1017:

With default window size, 854x480:

RealRTTV:
check out `mixin/screen/MinecraftMixin.java` L39. If you think there's a better way to recache gui scale, then I'm all ears, but currently it recaches on resisze & screen change

### DIFF
--- a/src/main/java/net/earthcomputer/clientcommands/interfaces/IScreenSafeZone.java
+++ b/src/main/java/net/earthcomputer/clientcommands/interfaces/IScreenSafeZone.java
@@ -1,0 +1,7 @@
+package net.earthcomputer.clientcommands.interfaces;
+
+public interface IScreenSafeZone {
+    int getSafeZoneWidth();
+
+    int getSafeZoneHeight();
+}
--- a/src/main/java/net/earthcomputer/clientcommands/mixin/screen/MinecraftMixin.java
+++ b/src/main/java/net/earthcomputer/clientcommands/mixin/screen/MinecraftMixin.java
@@ -7,6 +7,7 @@ import net.minecraft.client.Minecraft;
 import net.minecraft.client.Options;
 import net.minecraft.client.gui.screens.Screen;
 import org.jspecify.annotations.Nullable;
+import org.objectweb.asm.Opcodes;
 import org.spongepowered.asm.mixin.Final;
 import org.spongepowered.asm.mixin.Mixin;
 import org.spongepowered.asm.mixin.Shadow;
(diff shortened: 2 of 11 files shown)


src/main/java/net/earthcomputer/clientcommands/mixin/screen/WindowMixin.java
```diff
@@ -0,0 +1,29 @@
...
+import com.llamalad7.mixinextras.injector.ModifyExpressionValue;
+import com.mojang.blaze3d.platform.Window;
+import net.earthcomputer.clientcommands.interfaces.IScreenSafeZone;
+import net.minecraft.client.Minecraft;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.injection.At;
+
+@Mixin(Window.class)
+public class WindowMixin {
+    @ModifyExpressionValue(method = "calculateScale", at = @At(value = "CONSTANT", args = "intValue=320"))
```
Earthcomputer: Are there no named constants anywhere in the code for this?

src/main/java/net/earthcomputer/clientcommands/interfaces/IScreenSafeZone.java
```diff
@@ -0,0 +1,7 @@
+package net.earthcomputer.clientcommands.interfaces;
+
+public interface IScreenSafeZone {
```
Earthcomputer: Idk about this name, is there an official name for the concept in the Minecraft code? At the very least it should be `ISafeZoneScreen`

RealRTTV: nope, no name sadly, i think "safe zone" is the exact concept tho so we'll go with that

src/main/java/net/earthcomputer/clientcommands/mixin/screen/MinecraftMixin.java
```diff
@@ -0,0 +1,41 @@
...
+    @Shadow
+    @Final
+    public Options options;
+
+    @Shadow
+    public abstract boolean isEnforceUnicode();
+
+    @Definition(id = "screen", field = "Lnet/minecraft/client/Minecraft;screen:Lnet/minecraft/client/gui/screens/Screen;")
+    @Expression("this.screen = ?")
+    @Inject(method = "setScreen", at = @At(value = "MIXINEXTRAS:EXPRESSION", shift = At.Shift.AFTER))
```
Earthcomputer: This could use the `FIELD` injection point which is simpler than `MIXINEXTRAS:EXPRESSION`

Earthcomputer (changes requested):

src/main/java/net/earthcomputer/clientcommands/mixin/screen/MinecraftMixin.java
RealRTTV: fair

src/main/java/net/earthcomputer/clientcommands/mixin/screen/MinecraftMixin.java
```diff
@@ -0,0 +1,40 @@
+package net.earthcomputer.clientcommands.mixin.screen;
+
+import com.llamalad7.mixinextras.expression.Definition;
+import com.llamalad7.mixinextras.expression.Expression;
```
Earthcomputer: Unused imports

Earthcomputer (changes requested):
Your screen mixins are missing a `package-info.java`

Earthcomputer (approved):

PR #793, opened 2026-05-17
fix #280

### LINKED ISSUE #280: /crelog doesn't work with Fabric server w/ Fabric API
`/crelog` does not work (doesn't reconnect you) if Fabric API is present on the server. Nothing in log. Using Fabric API v. 0.35.0+1.17
Sorry for changing the title so much, you can see my thought process lol

RealRTTV:
v2 cause i didn't name branches properly

### DIFF
--- a/src/main/resources/clientcommands.aw
+++ b/src/main/resources/clientcommands.aw
@@ -48,7 +48,6 @@ accessible field net/minecraft/client/renderer/ShaderManager$CompilationCache co
 
 # crelog
 accessible field net/minecraft/client/gui/screens/DisconnectedScreen parent Lnet/minecraft/client/gui/screens/Screen;
-accessible field net/minecraft/client/gui/screens/DisconnectedScreen details Lnet/minecraft/network/DisconnectionDetails;
 
 # Game Options
 accessible field net/minecraft/client/OptionInstance value Ljava/lang/Object;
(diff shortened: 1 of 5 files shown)


src/main/resources/clientcommands.aw
```diff
@@ -46,6 +46,10 @@ accessible method net/minecraft/client/renderer/GameRenderer setPostEffect (Lnet
 accessible field net/minecraft/client/renderer/ShaderManager compilationCache Lnet/minecraft/client/renderer/ShaderManager$CompilationCache;
 accessible field net/minecraft/client/renderer/ShaderManager$CompilationCache configs Lnet/minecraft/client/renderer/ShaderManager$Configs;
 
+# crelog
+accessible field net/minecraft/client/gui/screens/DisconnectedScreen parent Lnet/minecraft/client/gui/screens/Screen;
+accessible field net/minecraft/client/gui/screens/DisconnectedScreen details Lnet/minecraft/network/DisconnectionDetails;
+
```
Earthcomputer, line 51: Are these still necessary?

RealRTTV: `parent` is required since otherwise it uses `mc.screen` for the parent sceen, meaning multiple relogs create a chain of `DisconnectedScreen` and escaping each one means going back a relog; not the intended behavior.

I could theoretically move the `parent` parameter in Relogger:107 and change it from `mc.screen` to an argument supplied on `loginToDedicatedServer` but I'd still need to get the parent of the `DisconnectedScreen` to put as the parent of the `ConnectScreen` when relogging -- which wouldn't solve the problem. IMO the current version is better because it doesn't have a parameter to loginToDedicatedServer for the screen to escape to afterwards. Mechanically it's pretty much identical to set the screen to the parent inside the `onFailedRelog` method instead of passing the parent screen to the `loginToDedicatedServer` function, it also makes that function more readable as it's logic is slightly simpler -- same goes for onFailedRelog.

`details` is barely not required anymore.

RealRTTV: i'd like to have details for safety on the callsite being at the correct time even though it doesn't need to be there. just for stability in the code.

your pick tho

Earthcomputer (approved):

PR #802, opened 2026-07-07
Allow copying chat lines when translating using /ctranslate

xpple:


src/main/java/net/earthcomputer/clientcommands/command/TranslateCommand.java
```diff
@@ -25,6 +25,8 @@
 
 public class TranslateCommand {
 
+    public static final String LITERAL = "ctranslate";
```
Earthcomputer: Probably better named `COMMAND_NAME` or something

src/main/java/net/earthcomputer/clientcommands/mixin/commands/translate/ChatComponent$1Mixin.java
```diff
@@ -0,0 +1,55 @@
...
+import net.minecraft.client.gui.screens.ChatScreen;
+import net.minecraft.client.multiplayer.chat.GuiMessage;
+import net.minecraft.client.renderer.state.gui.GuiTextRenderState;
+import net.minecraft.util.ARGB;
+import net.minecraft.util.FormattedCharSequence;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.injection.At;
+
+@Mixin(targets = "net/minecraft/client/gui/components/ChatComponent$1")
+public abstract class ChatComponent$1Mixin {
```
Earthcomputer: The `$` symbol is reserved for internal usage in Java, I know some in the modding community use it anyway but in this project we like to use `_` instead.

src/main/java/net/earthcomputer/clientcommands/mixin/commands/translate/ChatComponent$1Mixin.java
```diff
@@ -0,0 +1,55 @@
...
+    @WrapOperation(method = "accept", at = @At(value = "INVOKE", target = "Lnet/minecraft/client/gui/components/ChatComponent$ChatGraphicsAccess;handleMessage(IFLnet/minecraft/util/FormattedCharSequence;)Z"))
+    private boolean copyText(ChatComponent.ChatGraphicsAccess instance, int textTop, float opacity, FormattedCharSequence text, Operation<Boolean> original, GuiMessage.Line line) {
+        if (shouldCallOriginal(instance, textTop, text, line)) {
+            return original.call(instance, textTop, opacity, text);
+        }
+        // original.call() always returns false
```
Earthcomputer: I'm not sure at a glance what this means in context here

src/main/java/net/earthcomputer/clientcommands/mixin/commands/translate/ChatComponent$1Mixin.java
```diff
@@ -0,0 +1,55 @@
...
+        return false;
+    }
+
+    private static boolean shouldCallOriginal(ChatComponent.ChatGraphicsAccess instance, int textTop, FormattedCharSequence text, GuiMessage.Line line) {
```
Earthcomputer: This method should probably have a better name, especially since it has side effects. Maybe `tryCopy`, inverting the return value? Also, this method should be `@Unique`

src/main/java/net/earthcomputer/clientcommands/mixin/commands/translate/ChatScreenMixin.java
```diff
@@ -0,0 +1,34 @@
...
+    @Unique
+    private boolean isTranslating = false;
+
+    protected ChatScreenMixin(Component title) {
+        super(title);
+    }
+
+    @Inject(method = "onEdited", at = @At("TAIL"))
+    private void onEdited(String value, CallbackInfo ci) {
+        this.isTranslating = value.startsWith(Commands.COMMAND_PREFIX + TranslateCommand.LITERAL);
```
Earthcomputer: Storing this `isTranslated` boolean and keeping it up-to-date seems like unnecessary complexity and opportunity to get out of sync after a port to a new Minecraft version. May as well get the text of the chat box and check if it starts with the translate command inside of `tryCopy`

Earthcomputer (changes requested):

src/main/java/net/earthcomputer/clientcommands/mixin/commands/translate/ChatComponent$1Mixin.java
xpple: I don't see an example of that but sure.

src/main/java/net/earthcomputer/clientcommands/mixin/commands/translate/ChatScreenMixin.java
xpple: It's kept up to date in `onEdited`, which is kind of the purpose of that method (hook for changes). It cannot really get out of sync because the Mixin injection would fail. Checking it in `tryCopy` would also be a bit less efficient.

Earthcomputer: `tryCopy` doesn't get called that often anyway, just checking there is way simpler and avoids us having to maintain extra state

Earthcomputer (approved):

PR #805, opened 2026-07-30
Disable debug variable printing

SemmieDev:
Removes a left over `print = true`, which was dumping a bunch of info into the logs and aborts the injection.


Earthcomputer (approved):
Oops

